PR #2, opened 2021-07-09
Composer init

kpodsiad:
Implementation of [compose phase](https://yaml.org/spec/1.2/spec.html#id2762107).  
The composer task is to produce a node graph (think of it as a JSON AST) from a stream of events. At now it's a very straightforward implementation that doesn't take into account any metadata information (because Tokens don't have it too), but this will change with improvements of the parser stage.  
Also, error handling is not the best one, but I think it's good enough for the first iteration.

### DIFF
--- a/src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
+++ b/src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
@@ -25,28 +25,23 @@ object ComposerImpl extends Composer with NodeTransform:
     yield node
 
   override def fromEvents(events: List[Event]): Either[YamlError, Node] =
-    events match {
+    events match
       case Nil =>
         Left(YamlError("No events available"))
       case _ =>
         val (node, _) = composeNode(events)
         node
-    }
 
-  private def composeNode(events: List[Event]): ComposeResult = {
-    events match {
+  private def composeNode(events: List[Event]): ComposeResult =
+    events match
       case head :: tail =>
-        head match {
+        head match
           case StreamStart | DocumentStart => composeNode(tail)
           case SequenceStart               => composeSequenceNode(tail)
           case MappingStart                => composeMappingNode(tail)
           case s: Scalar                   => (composeScalarNode(s), tail)
-          case _                           => (Left(YamlError("No events available")), events)
-        }
-      case Nil =>
-        (Left(YamlError("No events available")), events)
-    }
-  }
+          case event                       => (Left(YamlError(s"Unexpected event $event")), events)
+      case Nil => (Left(YamlError("No events available")), Nil)
 
   private def composeSequenceNode(events: List[Event]): ComposeResult = {
     @tailrec
--- a/src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
+++ b/src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
@@ -36,10 +36,10 @@ object ComposerImpl extends Composer with NodeTransform:
       case head :: tail =>
         head match
           case Event.StreamStart | Event.DocumentStart => composeNode(tail)
-          case Event.SequenceStart               => composeSequenceNode(tail)
-          case Event.MappingStart                => composeMappingNode(tail)
-          case s: Event.Scalar                   => (composeScalarNode(s), tail)
-          case event                       => (Left(YamlError(s"Unexpected event $event")), events)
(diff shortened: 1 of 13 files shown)


src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
```diff
@@ -0,0 +1,98 @@
...
+
+trait Composer:
+  def compose(yaml: String): Either[YamlError, Node]
+  def compose(reader: YamlReader): Either[YamlError, Node]
+
+object ComposerImpl extends Composer with NodeTransform:
+
+  type ComposeResult = (Either[YamlError, Node], List[Event])
+
+  override def compose(yaml: String): Either[YamlError, Node] = compose(StringYamlReader(yaml))
```
lwronski, line 18: I propose to change the name StringYamlReader to YamlReader. 
The name StringYamlReader it would make sense only if we supported binary serialization.

```suggestion
  override def compose(yaml: String): Either[YamlError, Node] = compose(YamlReader(yaml))
```

src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
```diff
@@ -0,0 +1,98 @@
...
+      case Nil => (Left(YamlError("No events available")), Nil)
+
+  private def composeSequenceNode(events: List[Event]): ComposeResult = {
+    @tailrec
+    def parseChildren(
+        events: List[Event],
+        children: List[Node]
+    ): (Either[YamlError, List[Node]], List[Event]) = {
+      events match
+        case Nil                 => (Left(YamlError("No events available")), Nil)
```
lwronski, line 52:
```suggestion
        case Nil                 => (Left(YamlError("Not found SequenceEvent for sequence)), Nil)
```

src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
```diff
@@ -0,0 +1,98 @@
...
+        case Event.SequenceEnd :: tail => (Right(children), tail)
+        case _ =>
+          val (node, rest) = composeNode(events)
+          node match
+            case Right(value) => parseChildren(rest, children :+ value)
+            case Left(err)    => (Left(err), rest)
```
lwronski, line 58: In any case of error we should skip any further parsing, so why is `rest` returned here?

src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private def composeMappingNode(events: List[Event]): ComposeResult = {
+    @tailrec
+    def parseMappings(
+        events: List[Event],
+        mappings: List[Node.KeyValueNode]
+    ): (Either[YamlError, List[Node.KeyValueNode]], List[Event]) = {
+      events match
+        case Nil                => (Left(YamlError("No events available")), Nil)
```
lwronski, line 73:
```suggestion
        case Nil                => (Left(YamlError("Not found MappingEnd event for mapping")), Nil)
```

src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
kpodsiad: Yeah, that could be `Nil`. It' not ideal that the result is Tuple[Either, List] because for Left there is no sense in returning a List of remaining events but this could be improved later

src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
kpodsiad: What about trait `YamlReader`? After renaming there will be a name conflict between them.

KacperFKorban (approved):
Looks great!

src/main/scala/org/virtuslab/internal/load/compose/Composer.scala
kpodsiad: I'll leave it as it is because you changed that name already in your branch ;)

kpodsiad:
> One note for possible improvement (which I think already said on slack, but this time in a 'weaker' version): changing `ComposeResult` to `Either[YamlError, (Node, List[Event])]` should allow for replacing some explicit matches on `Either` with `map`/`flatMap`/`for`.

I read it on Friday then forgot about it 😅.  Anyway, thanks for the _Weaker_ version, the code is more concise now.

PR #4, opened 2021-07-12
Construct typeclass derivation

kpodsiad:
Implementation of [construct stage](https://yaml.org/spec/1.2/spec.html#id2762107).  It's a poc and feedback will be extremely welcome.

1. I don't have experience with Magnolia so instead of learning it, I wrote derivation using Mirrors (which I played with few times). I'm not happy how derivation for `Mirror.SumOf` works, but IMO it's a good starting point.
2. I decided to introduce two typeclasses:
    * `Construct` which is responsible for constructing an instance of `T` from `Node`.
    * `Der` (rather placeholder than a final name) which is almost a `String => T` function.  

Although it seems reasonable to split `Node => String => T` into two phases to achieve single responsibility and easier customization (it's harder to create a new Construct than Der for the end-user IMO), I'm not convinced of the current implementation.


src/main/scala/org/virtuslab/internal/load/construct/Construct.scala
```diff
@@ -2,15 +2,80 @@ package org.virtuslab.internal.load.construct
...
+      override def construct(node: Node): Either[YamlError, T] = {
+        def loop(
+            constructs: List[Construct[T]],
+            acc: Either[YamlError, T]
+        ): Either[YamlError, T] =
+          constructs match
+            case head :: tail => if acc.isLeft then loop(tail, head.construct(node)) else acc
+            case Nil          => acc
+
+        loop(instances.tail, instances.head.construct(node))
```
KacperFKorban, line 72: Maybe using [collectFirst](https://dotty.epfl.ch/api/scala/collection/IterableOnceOps.html) or sth similar may make the implementation easier.

src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
```diff
@@ -3,28 +3,17 @@ package org.virtuslab.yaml
...
-import org.virtuslab.internal.load.construct.Construct
-import org.virtuslab.internal.load.compose.Node
 import org.virtuslab.internal.YamlError
-import org.virtuslab.internal.load.compose.ComposerImpl
+import org.virtuslab.internal.load.{YamlReader, StringYamlReader}
+import org.virtuslab.internal.load.parse.ParserImpl
+import org.virtuslab.internal.load.construct.{Der}
+import org.virtuslab.internal.load.compose.{Node, ComposerImpl}
 
 trait YamlDecoder[T: Mirror.Of]:
```
KacperFKorban, line 12: IMHO having `Mirror` constraints on type constructor isn't the best practice, since it disallows providing own (manual) implementations.

lbialy:
I'll leave my review and opinion later today.

src/main/scala/org/virtuslab/internal/load/construct/Construct.scala
kpodsiad: I ended using `LazyList` and `collectFirst` and it's for sure more readable.

src/main/scala/org/virtuslab/yaml/yaml.scala
```diff
@@ -7,8 +7,8 @@ import org.virtuslab.internal.YamlError
 
 import scala.deriving._
 
-def deriveEncoder[T](using m: Mirror.Of[T]) = YamlEncoder.derived
-def deriveDecoder[T](using m: Mirror.Of[T]) = YamlDecoder.derived
+// def deriveEncoder[T](using m: Mirror.Of[T]) = YamlEncoder.derived
+// def deriveDecoder[T](using m: Mirror.Of[T]) = YamlDecoder.derived
```
lwronski, line 11: If you commented any code,  please write a short explanatory comment or just remove code.

src/main/scala/org/virtuslab/yaml/yaml.scala
```diff
@@ -1,17 +1,16 @@
...
-
-def parse(yaml: String): Either[YamlError, Node] = ComposerImpl.compose(yaml)
-extension (str: String)
-  def fromYaml[T](using decoder: YamlDecoder[T]): Either[Any, T] = decoder.from(str)
+inline def deriveEncoder[T](using m: Mirror.Of[T]): YamlEncoder[T] = YamlEncoder.derived[T]
+inline def deriveDecoder[T](using m: Mirror.Of[T]): Construct[T]   = Construct.derived[T]
 
-extension [T](t: T) def asYaml(using encoder: YamlEncoder[T]): String = encoder.toYaml(t)
+extension (str: String) def as[T](using c: Construct[T]): Either[YamlError, T] = Util.as(str)
+extension [T](t: T) def asYaml(using encoder: YamlEncoder[T]): String          = Util.asYaml(t)
```
kpodsiad, line 16: `munit` had a problem with extension methods written at top-level so I moved it to the `Util` object. It's a little boilerplate to repeat `extension (str: String) def as[T]...` 2 times, but at least there is only one implementation of that method

kpodsiad:
After discussion we decided to drop `Der` - `Construct` should be enough and with proper factory methods (like existing `apply` it could be convenient to use.

I wonder how those factory methods should look like in terms of error handling. Should failure be explicite included in type like `Try` or `Either` or, for simplicity sake, just `T` and try/catch in factory would be enough?
```scala
object Construct:
  def primitive[T](read: String => T): Construct[T] = ???
  def primitive[T](read: String => Either[Throwable, T]): Construct[T] = ???
  def primitive[T](read: String => Try[T]): Construct[T] = ???
```

PR #5, opened 2021-07-13
Implement of writer stage

lwronski:
Implementation of writer stage. It's a straightforward way to parsing case class to yaml.

For now, we only support simple case classes without nested objects.


kpodsiad (approved):
LGTM

PR #9, opened 2021-07-16
Rename Construct & YamlWriter into YamlDecoder & YamlEncoder

kpodsiad:
Also, I created type aliases for some type and placed them in `Yaml` object. Thanks to that, `import org.virtuslab.Yaml.*` imports everything which is needed to use the library by end-user.


lwronski (approved):
LGTM

Issue #13, opened 2021-07-16
Support explicit document end (three dots)


kpodsiad:
https://yaml.org/spec/1.2/spec.html#id2760395

PR #15, opened 2021-07-17
Parse DocumentEnd

kpodsiad:
Implements https://github.com/VirtusLab/scala-yaml/issues/13

### DIFF
--- a/src/main/scala/org/virtuslab/internal/load/reader/ReaderState.scala
+++ b/src/main/scala/org/virtuslab/internal/load/reader/ReaderState.scala
@@ -1,11 +1,8 @@
 package org.virtuslab.internal.load.reader
 
-sealed trait ReaderState:
-  def indent: Int
+sealed trait ReaderState
 case object ReaderState:
-  case object Stream extends ReaderState:
-    override val indent = 0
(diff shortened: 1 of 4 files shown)


src/main/scala/org/virtuslab/internal/load/reader/ReaderState.scala
```diff
@@ -2,9 +2,10 @@ package org.virtuslab.internal.load.reader
 
 sealed trait ReaderState:
   def indent: Int
-
 case object ReaderState:
-  case class Stream(indent: Int)   extends ReaderState
-  case class Document(indent: Int) extends ReaderState
+  case object Stream extends ReaderState:
+    override val indent = 0
```
lwronski, line 7: If you don't use val indent in Stream and Document, maybe you should delete them.

lwronski (commented):
LGTM

src/main/scala/org/virtuslab/internal/load/reader/ReaderState.scala
kpodsiad: Yeah, you're right.

PR #16, opened 2021-07-17
Parser error handling

kpodsiad:
Closes https://github.com/VirtusLab/scala-yaml/issues/12

### DIFF
--- a/src/main/scala/org/virtuslab/internal/load/parse/ParserImpl.scala
+++ b/src/main/scala/org/virtuslab/internal/load/parse/ParserImpl.scala
@@ -1,6 +1,7 @@
 package org.virtuslab.internal.load.parse
 
 import org.virtuslab.internal.YamlError
+import org.virtuslab.internal.ParseError
 import org.virtuslab.internal.load.reader.Reader
 import org.virtuslab.internal.load.reader.token.ScalarStyle
 import org.virtuslab.internal.load.reader.token.Token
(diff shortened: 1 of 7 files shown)


src/main/scala/org/virtuslab/internal/load/parse/ParserImpl.scala
```diff
@@ -39,68 +39,96 @@ private enum Production:
...
+          else Right(acc :+ event)
+        case Left(err) => Left(err)
     }
-    Right(loop(Nil, List(ParseStreamStart)))
+    loop(Nil, List(ParseStreamStart))
 
-  def getNextEvent(in: Reader, stack: List[Production]): (Event, List[Production]) = {
+  type EventResult = Either[YamlError, (Event, List[Production])]
+
+  private def error(expected: String, got: Token): EventResult = Left(
```
lwronski, line 58: We should separate the errors logic from the parsing. We should cover error parsing with new type error. 
In new type error, you can create helper method for parsing, such as 

`EventError.from(expected: Token, got: Token): EventResult`

lwronski (commented):
We should separate errors logic from parsing.

src/main/scala/org/virtuslab/internal/load/parse/ParserImpl.scala
kpodsiad: Should we change `YamlError` to sealed trait and then, `EventError` will extend it?

lwronski: Yes, good idea

lwronski (approved):
LGTM

PR #21, opened 2021-07-20
YamlEncoder reimplementation

### LINKED ISSUE #18: Support nested types in YamlEncoder
For example, encoding `Person` would produce malformed output.
```scala
  case class Address(city: String, street: String) derives YamlEncoder
  case class Person(name: String, address: Address) derives YamlEncoder
```

kpodsiad:
closes #18
[YAML dump](https://yaml.org/spec/1.2/spec.html#id2762313) (serialize) phase consists of 3 stages and this PR implements each of them


lwronski (approved):
LGTM

PR #23, opened 2021-07-20
Implement YamlDecoder instances for primitive types

lwronski:


src/main/scala/org/virtuslab/internal/load/construct/YamlDecoder.scala
```diff
@@ -9,29 +8,110 @@ import scala.compiletime._
...
 sealed trait YamlDecoder[T]:
-  def construct(node: Node): Either[YamlError, T]
+  def construct(node: Node): Either[ConstructError, T]
 
 object YamlDecoder:
-  def apply[T](pf: PartialFunction[Node, Either[Throwable, T]]): YamlDecoder[T] =
+  def apply[T](pf: PartialFunction[Node, Either[ConstructError, T]]): YamlDecoder[T] =
     new YamlDecoder[T] {
-      override def construct(node: Node): Either[YamlError, T] =
+      override def construct(node: Node): Either[ConstructError, T] =
```
kpodsiad, line 19: Since `ConstructError` extends `YamlError`, it isn't necessary to change this type, is it?

kpodsiad, line 19: Since ConstructError extends YamlError, it isn't necessary to change this type, is it?

src/main/scala/org/virtuslab/internal/load/construct/YamlDecoder.scala
```diff
@@ -9,29 +8,110 @@ import scala.compiletime._
 import scala.compiletime.summonFrom
 import scala.deriving._
 import scala.util.Try
+import scala.language.implicitConversions
```
kpodsiad, line 11: I would prefer explicit conversion rather than an implicit one because it's.. more explicit and less magical

src/main/scala/org/virtuslab/internal/YamlError.scala
```diff
@@ -2,7 +2,9 @@ package org.virtuslab.internal
 
 import org.virtuslab.internal.load.reader.token.Token
 
-sealed trait YamlError
+sealed trait YamlError extends Throwable:
```
kpodsiad, line 5: `extends Throwable`?

src/main/scala/org/virtuslab/internal/load/construct/YamlDecoder.scala
```diff
@@ -9,29 +8,110 @@ import scala.compiletime._
...
-  def apply[T](pf: PartialFunction[Node, Either[Throwable, T]]): YamlDecoder[T] =
+  def apply[T](pf: PartialFunction[Node, Either[ConstructError, T]]): YamlDecoder[T] =
     new YamlDecoder[T] {
-      override def construct(node: Node): Either[YamlError, T] =
+      override def construct(node: Node): Either[ConstructError, T] =
         if pf.isDefinedAt(node) then
           pf(node) match
             case Right(value)    => Right(value)
             case Left(throwable) => Left(ConstructError(throwable.toString))
         else Left(ConstructError(s"Could't create Construct instance for $node"))
```
kpodsiad, line 24: This apply doesn't work correctly because pf already returs `ConstructError` in the result (`pf: PartialFunction[Node, Either[ConstructError, T]])`)
Maybe it's worth having secondary apply for `Throwable` and get rid of implicit conversion?

kpodsiad (changes requested):

src/main/scala/org/virtuslab/internal/load/construct/YamlDecoder.scala
lwronski: On the one hand yes, but in YamlDecoder we expect throwing only ConstructError, we don't want for example parse error here.

src/main/scala/org/virtuslab/internal/load/construct/YamlDecoder.scala
lwronski: Good point, it should be removed immediately :)

src/main/scala/org/virtuslab/internal/load/construct/YamlDecoder.scala
lwronski: You are quite right, but it's works because `pf(node)` returned `Either[ConstructError, T]`.

so there was some redundancy here
```
pf(node) match
            case Right(value)    => Right(value)
            case Left(throwable) => Left(ConstructError(throwable.toString))

```
that could be rewritten to `pf(node)`

src/main/scala/org/virtuslab/internal/YamlError.scala
lwronski: Good point, it should be removed immediately :)

src/main/scala/org/virtuslab/internal/load/construct/YamlDecoder.scala
lwronski: Sure, I removed implicit conversion

kpodsiad (approved):
LGTM

PR #26, opened 2021-07-21
Parse empty flow sequence and add support for double quote key in mapping

lwronski:
* Parsing multiline scalar with single quote style 
* Allow to parsing mapping with scalar double quote style as key
* Parsing empty flow sequence

### DIFF
--- a/src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
+++ b/src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
@@ -17,9 +17,10 @@ private enum Production:
   case ParseMappingEnd
   case ParseFlowMappingEnd
   case ParseSequenceStart
+  case ParseFlowSequenceStart
+  case ParseFlowSequenceEnd
   case ParseSequenceEnd
   case ParseNode
-  case ParseNodeOpt
   case ParseKey
   case ParseValue
   case ParseOptKey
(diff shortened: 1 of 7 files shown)


src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
```diff
@@ -33,14 +34,15 @@ private enum Production:
  * ParseDocumentStart    ::= <ParseNode> ParseDocumentEnd
  * ParseDocumentStartOpt ::= epsilon | <ParseDocumentStart> <ParseDocumentStartOpt>
  * ParseNode             ::= <ParseMappingStart> | <ParseFlowMappingStart> | <ParseSequenceStart> | ParseScalar
+ * ParseNodeOpt          ::= epsilon | <ParseMappingStart> | <ParseFlowMappingStart> | <ParseSequenceStart> | ParseScalar
  * ParseMappingStart     ::= <ParseKey> ParseMappingEnd
  * ParseFlowMappingStart ::= <ParseFlowMappingEntry> ParseFlowMappingEnd
  * ParseFlowMappingEntry ::= <ParseScalar> | <ParseFlowMappingEntry> <ParseFlowMappingEnd> <ParseNode>
- * ParseSequenceStart    ::= <ParseNode> <ParseSequenceEntryOpt> ParseSequenceEnd
+ * ParseSequenceStart    ::= <ParseNodeOpt> <ParseSequenceEntryOpt> ParseSequenceEnd
```
kpodsiad, line 41: Grammar is incorrect because `FlowSequence` can be empty but `Block Sequence` can't. These two kinds of sequences should be distiguished.

src/main/scala/org/virtuslab/yaml/internal/load/reader/YamlReader.scala
```diff
@@ -98,29 +102,37 @@ class YamlReader(in: CharSequence) extends Reader {
     val foldedIndent = indent
     skipUntilNextIndent(foldedIndent)
 
+    def chompedEmptyLines() =
+      while (peek() == Some('\n')) {
+        skipCharacter()
+        sb.append("\\n")
+      }
+
+      skipUntilNextIndent(foldedIndent)
```
kpodsiad, line 111: in [spec](https://yaml.org/spec/1.2/spec.html#id2794534) `chomping` _controls how final line breaks and trailing empty lines are interpreted_. For me `chomping` is also a very enigmatic word, I think that, for example, `eatNewlines` would be clearer

src/test/scala/org/virtuslab/yaml/internal/load/parse/ScalarSpec.scala
```diff
@@ -32,6 +32,26 @@ class ScalarSpec extends munit.FunSuite:
...
+         |""".stripMargin
+
+    val reader = YamlReader(yaml)
+    val events = ParserImpl.getEvents(reader)
+
+    val expectedEvents = Right(
+      List(
+        StreamStart,
+        DocumentStart(),
+        Scalar("mnt\\\\#dd", ScalarStyle.Plain),
```
kpodsiad, line 47: If you use `raw` string interpolator, you won't have to escape `\`

kpodsiad (changes requested):
I think FlowSequence should have its own grammar rule

src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
lwronski: Right, it was mistake, so I added new production state `ParseFlowSequenceStart` which allows to empty sequences.

kpodsiad (approved):
LGTM

PR #28, opened 2021-07-21
YamlCodec

kpodsiad:
Wrap YamlEncoder & YamlDecoder into YamlCodec.
Use `inline def derived` instead of `inline given derived` in order to avoid automatic typeclass generation


lwronski (approved):
LGTM

PR #31, opened 2021-07-22
Configure nicely scaladoc

romanowski:


build.sbt
```diff
@@ -31,13 +31,23 @@ lazy val munit: Seq[Setting[_]] = Seq(
...
+  "-project", "Scala-yaml",
+  "-siteroot", "docs",
+  "-project-version", version.value,
+  "-project-logo", "docs/logo.svg",
+  "-social-links:" +
+    "github::https://github.com/VirtusLab/scala-yaml," +
+    "twitter::https://twitter.com/VirtusLab",
+  "-project-footer", s"Copyright (c) 2021, VirtusLab",
+  "-source-links:github://VirtusLab/scala-yaml",
+  "-revision", "master"
```
kpodsiad, line 43: According to [docs](https://dotty.epfl.ch/docs/usage/scaladoc/settings.html) `-siteroot docs` and `-revision master` are default options so they can be omitted.

kpodsiad (approved):

build.sbt
romanowski: Docs are misleading since we need them - we will work on scaladoc docs

PR #43, opened 2021-07-28
Extract Reader from Scanner

kpodsiad:


src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
```diff
@@ -260,41 +262,27 @@ private[yaml] class YamlReader(in: CharSequence) extends Reader {
...
     indent = 0
-    while (peek() == Some(' ') && indent < indentBlock) {
+    while (in.peek() == Some(' ') && indent < indentBlock) {
       indent += 1
-      skipCharacter()
+      in.skipCharacter()
     }
 
+  private def skipComment(): Unit =
+    while !(in.peek() == Some('\n') || in.peek() == Some('\r')) do in.skipCharacter()
```
lwronski, line 287: I think, that we should move this logic to new method `in.isNextNewLine`

lwronski (approved):
LGTN

src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
kpodsiad: I agree. What's more `\r\n` are two characters and we are skipping only one :/
I'll leave it as is because it needs more refactoring than just `in.isNextNewLine`

PR #45, opened 2021-07-31
Add drafter to workflows

kpodsiad:
[Drafter readme](https://github.com/release-drafter/release-drafter#readme)


.github/release-drafter-config.yml
```diff
@@ -0,0 +1,48 @@
...
+  - title: '🚀 Features'
+    labels:
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'bug'
+autolabeler:
+  - label: 'documentation'
+    files:
+      - '*.md'
```
lwronski, line 13: It's a good idea to include all the files in the docs directory.

.github/release-drafter-config.yml
```diff
@@ -0,0 +1,48 @@
...
+    branch:
+      - '/docs{0,1}\/.+/'
```
lwronski, line 15: I don't think I need this option, we don't have this naming for branch

.github/release-drafter-config.yml
```diff
@@ -0,0 +1,48 @@
...
+  - label: 'bug'
+    branch:
+      - '/fix\/.+/'
```
lwronski, line 18: How about the naming branch with prefix `bugfix`?

lwronski (commented):
We should standardize branch names, my proposition
- branch with bugs `'/bugfix\/.+/'`
- branch with feature `'/feature\/.+/'`

.github/release-drafter-config.yml
kpodsiad: I think that for the autolabelel purpose `.md` plus `.svg` should be good enough. Currently, our docs consist of only md and svg files

.github/release-drafter-config.yml
kpodsiad: Yeah, true. For docs, file matcher should be enough

.github/release-drafter-config.yml
lwronski: Hope you will remember that :) 
Yes, it is good enough for now, but later there will be files with other extension.

lwronski (approved):
When CI is finish successfully, you can merge.

kpodsiad:
> When CI is finish successfully, you can merge.

According to the https://github.com/release-drafter/release-drafter/issues/398 config file should be on the default branch, otherwise workflow will fail. So I'm gonna transplant the config file into main and then check if workflow passes.

PR #49, opened 2021-08-02
Improve error messages

kpodsiad:
TODO:
- [x] add `Input` to the `Position` like it is in e.g [scalameta trees Position](https://github.com/scalameta/scalameta/blob/main/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Position.scala)
- [x] use `Position` to create error messages like the following. Maybe not all of them would be done in this PR but I want to implement at least one of them. 
```terminal
Error: <line>:<column> expected value in mapping, but got nothing instead
key: <nothing>
     ^^^^^^^^^
``` 
```terminal
Error: <line>:<column> expected int, got string instead
key: "string_value"
     ^^^^^^^^^^^^^^
``` 
- [x] My changes broke all the tests and I have to do sth to fix this.


yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
```diff
@@ -39,49 +41,53 @@ object ComposerImpl extends Composer:
     @tailrec
     def parseChildren(
         events: List[Event],
-        children: List[Node]
-    ): ComposeResult[List[Node]] = events match
+        children: List[Node],
+        startPos: Option[Position] = None
```
lwronski, line 45:
```suggestion
        pos: Option[Position] = None
```

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
```diff
@@ -18,7 +19,8 @@ trait Composer:
   def fromEvents(events: List[Event]): Either[YamlError, Node]
 
 object ComposerImpl extends Composer:
-  type ComposeResult[T] = Either[YamlError, (T, List[Event])]
+  type ComposeResult[T]        = Either[YamlError, (T, List[Event])]
+  type ComposeResultWithPos[T] = Either[YamlError, (T, Option[Position], List[Event])]
```
lwronski, line 23: Why do we need to separate into two types? 

`(T, Option[Position], List[Event])` it can wrapped to new type or case class. It will be more readable

lwronski (commented):
LGTM,  waiting for minor corrections

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
kpodsiad: Those types should be private. `ComposeResultWithPos` is used in tailrecs because they should return triple (parsed node, first child position, rest of the events). Thanks to that other methods don't bother with unused position parameter

lwronski (approved):
LGTM, squash commits before merge

PR #65, opened 2021-09-05
Include new lines for single quote style

lwronski:


yaml/shared/src/test/scala/org/virtuslab/yaml/internal/load/parse/ScalarSpec.scala
```diff
@@ -48,6 +48,33 @@ class ScalarSpec extends BaseParseSuite:
     assertEventsEquals(events, expectedEvents)
   }
 
+  test("should parse duble quote scalar value with multiline") {
```
kpodsiad, line 51:
```suggestion
  test("should parse single quote scalar value with multiline") {
```

kpodsiad (approved):

PR #66, opened 2021-09-06
Parse new lines for plain scalar style

lwronski:


yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/ReaderState.scala
```diff
@@ -2,9 +2,11 @@ package org.virtuslab.yaml.internal.load.reader
...
 sealed trait ReaderState
 case object ReaderState:
-  case object Stream                     extends ReaderState
-  case object Document                   extends ReaderState
-  final case class Mapping(indent: Int)  extends ReaderState
-  final case class Sequence(indent: Int) extends ReaderState
-  case object FlowMapping                extends ReaderState
-  case object FlowSequence               extends ReaderState
+  case object Stream   extends ReaderState
+  case object Document extends ReaderState
```
kpodsiad, line 6: `Stream` is unused and and `Document` has only one usage. If we get rid off them, `ReaderStateWithIndent` won't be necessary

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
```diff
@@ -235,22 +231,27 @@ private[yaml] class Scanner(str: String) extends Tokenizer {
...
-              .peekNext() == Some('\r') =>
-          sb.result()
+        case Some(':') if in.isNextWhitespace                   => sb.result()
         case Some(char) if !ctx.isAllowedSpecialCharacter(char) => sb.result()
         case Some(' ') if in.peekNext() == Some('#')            => sb.result()
-        case Some('\n') | Some('\r') | None                     => sb.result()
+        case _ if in.isNewline =>
+          skipUntilNextChar()
+          sb.append(' ')
+          if (in.column > ctx.getIndentOfLatestCollection().getOrElse(Int.MaxValue)) readScalar()
```
kpodsiad, line 245:
```suggestion
          if (ctx.getIndentOfLatestCollection().exists(_ <= in.column)) readScalar()
```

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/ReaderState.scala
kpodsiad, line 6: `Stream` is unused and and `Document` has only one usage. If we get rid of them, `ReaderStateWithIndent` won't be necessary

kpodsiad (approved):

PR #70, opened 2021-09-09
Improve release workflow

kpodsiad:
Don't publish tests modules
Run CI workflow on published release


lwronski (approved):
LGTM, please, before merging, check if scala doc is generated correctly

Issue #69, opened 2021-09-09
Allow incremental processing of documents


kpodsiad:
According to the [spec](https://yaml.org/spec/1.2/) _YAML was designed to support incremental interfaces that include both input (“getNextEvent()”) and output (“sendNextEvent()”) one-pass interfaces_. 

However, the current implementation allows only batched processing of whole documents. This change would be also very handy when working with the official test suite because it has negative tests, for instance [this one](https://github.com/yaml/yaml-test-suite/blob/master/test/2CMS.tml). Those test cases contain a list of expected events before an error occurrence.

PR #71, opened 2021-09-14
Add incremental processing for parsed events

kpodsiad:
Implements `getNextEvent` mentioned in https://github.com/VirtusLab/scala-yaml/issues/69

### DIFF
--- a/yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/Parser.scala
+++ b/yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/Parser.scala
@@ -4,5 +4,4 @@ import org.virtuslab.yaml.YamlError
 import org.virtuslab.yaml.internal.load.reader.Tokenizer
 
 trait Parser:
-  def getEvents(): Either[YamlError, Seq[Event]]
   def getNextEvent(): Either[YamlError, Event]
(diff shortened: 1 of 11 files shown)


yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/Parser.scala
```diff
@@ -2,5 +2,7 @@ package org.virtuslab.yaml.internal.load.parse
 
 import org.virtuslab.yaml.YamlError
 import org.virtuslab.yaml.internal.load.reader.Tokenizer
+
 trait Parser:
-  def getEvents(in: Tokenizer): Either[YamlError, Seq[Event]]
+  def getEvents(): Either[YamlError, Seq[Event]]
```
lwronski, line 7: This means I can call getEvents and then getNextEvent. You should break it down into two traits.

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
```diff
@@ -49,202 +51,203 @@ private enum Production:
...
+        case Right(event) =>
+          if event != Event.StreamEnd then loop(event :: acc)
+          else Right((event :: acc).reverse)
         case Left(err) => Left(err)
     }
-    loop(Nil, List(ParseStreamStart))
+    loop(Nil)
 
-  type EventResult = Either[YamlError, (Event, List[Production])]
+  override def getNextEvent(): Either[YamlError, Event] =
```
lwronski, line 72: In any case, after the StreamEvent event, the parser will not return another event? I think we should use Option.

kpodsiad: I changed this behavior. Now, it's returning right with `Event.StreamEnd` value if there are no more events

lwronski (approved):
LGTM

Issue #68, opened 2021-09-08
Add API/dsl which will provide a convenient way to construct nodes


kpodsiad:
### Discussed in https://github.com/VirtusLab/scala-yaml/discussions/56

<div type='discussions-op-text'>

<sup>Originally posted by **pikinier20** August 19, 2021</sup>
- Add a DSL to construct Nodes

Examples:
```scala
yamlObj(
  "field1" -> true,
  "field2" -> 2,
  "field3" -> yamlObj(
    "field4" - "asd"
  )
)
```
</div>

PR #72, opened 2021-09-19
Add convenient way to create YAML nodes in code

kpodsiad:
~Aims to close https://github.com/VirtusLab/scala-yaml/issues/68~
Will be closed in a follow-up PR.

Current state: one can write sth like this
```scala
MappingNode(
      "hr"   -> ScalarNode("65"),
      "avg"  -> ScalarNode("0.278"),
      "blob" -> ScalarNode("nostradamus")
    )
```

I was thinking about implicit conversions that would allow more concise syntax:
```scala
MappingNode(
      "hr"   -> 65,
      "avg"  -> 0.278),
      "blob" -> "nostradamus",
      "bool_true" -> true,
      "bool_false" -> false,
    )
```
As we know, implicit conversions could be very tricky so I would prefer t place them at `org.virtuslab.yaml.syntax` to limit accidental import. In addition to import this package, user will have to import `import scala.language.implicitConversions` or set proper compiler option. Otherwise, he will get compiler warnings about usage of implicit conversion. 
I find this proposal as a good balance between convenience and user awareness of what he is doing. What do you think?

### DIFF
--- a/yaml/shared/src/main/scala/org/virtuslab/yaml/Node.scala
+++ b/yaml/shared/src/main/scala/org/virtuslab/yaml/Node.scala
@@ -1,6 +1,7 @@
 package org.virtuslab.yaml
 
 import org.virtuslab.yaml.Position
+import org.virtuslab.yaml.syntax.YamlPrimitive
 
 /**
   * ADT that corresponds to the YAML representation graph nodes https://yaml.org/spec/1.2/spec.html#id2764044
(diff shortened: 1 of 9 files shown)


yaml/shared/src/main/scala/org/virtuslab/yaml/Node.scala
```diff
@@ -9,16 +9,34 @@ sealed trait Node:
...
 object Node:
-  final case class ScalarNode(value: String, pos: Option[Position] = None) extends Node
+  final case class ScalarNode private[yaml] (value: String, pos: Option[Position] = None)
+      extends Node
 
-  final case class SequenceNode(nodes: Seq[Node], pos: Option[Position] = None) extends Node
-  object SequenceNode:
-    def apply(nodes: Node*): SequenceNode = SequenceNode(nodes, None)
+  object ScalarNode:
+    given c1: (String => ScalarNode)     = ScalarNode(_)
```
lwronski, line 16: This name is ambiguous, you don't have to define the given name here, it will be easier

lwronski (approved):
Good point, we should allow to use implicit conversions if the user imported special object from scala-yaml. We will add it in a separate PR.

Please, add at least one test for testing your implicit for nested yaml. For example
```
Sequence...
         Mapping...
         Mapping...
```

yaml/shared/src/main/scala/org/virtuslab/yaml/Node.scala
kpodsiad: Yeah, it can be anonymous given, couldn't agree more!

yaml/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
```diff
@@ -134,7 +134,7 @@ object YamlDecoder:
       override def construct(node: Node): Either[ConstructError, T] =
         node match
           case Node.MappingNode(mappings, _) =>
-            val valuesMap = mappings.map(e => e.key.value -> e.value).toMap
+            val valuesMap = mappings.map(e => e.key.asInstanceOf[ScalarNode].value -> e.value).toMap
```
lwronski, line 137: @kpodsiad What should happen if scalarNode doesn't appear there? 
throw an exception that such node isn't allowed?

kpodsiad: Yes, we should proper `ConstructError` because for product types key has to be an instance of ScalarNode

yaml/shared/src/main/scala/org/virtuslab/yaml/syntax/YamlPrimitive.scala
```diff
@@ -0,0 +1,23 @@
...
+  given Conversion[Node, YamlPrimitive]    = s => YamlPrimitive(s)
+
+  given conv[A, B](using
+      Conversion[A, YamlPrimitive],
+      Conversion[B, YamlPrimitive]
+  ): Conversion[(A, B), (YamlPrimitive, YamlPrimitive)] = (s, b) => (s, b)
+
+  given convMulti[A, B](using
+      Conversion[A, YamlPrimitive]
+  ): Conversion[(A, Node), (YamlPrimitive, YamlPrimitive)] = (s, b) => (s, YamlPrimitive(b))
```
kpodsiad, line 23: I think we could do it in more explicit way
```suggestion
  given [A, B](using
      c1: Conversion[A, YamlPrimitive],
      c2: Conversion[B, YamlPrimitive]
  ): Conversion[(A, B), (YamlPrimitive, YamlPrimitive)] = (a, b) => (c1(a), c2(b))

  given [A](using
      c: Conversion[A, YamlPrimitive]
  ): Conversion[(A, Node), (YamlPrimitive, YamlPrimitive)] = (a, node) => (c(a), YamlPrimitive(node))
```

yaml/shared/src/test/scala/org/virtuslab/yaml/internal/NodeCreationSuite.scala
```diff
@@ -0,0 +1,81 @@
...
+
+//  test("scalar-node") {
+//    val node: ScalarNode = "65"
+//
+//    assertEquals(
+//      node.asYaml,
+//      """|65
+//         |""".stripMargin
+//    )
+//  }
```
kpodsiad, line 50: can be deleted

kpodsiad (commented):
LGTM, cannot approve because I'm author of the PR 😅

PR #76, opened 2021-10-07
Simplify Tokenizer logic

kpodsiad:
Previously, `Tokenizer` contained too much logic and it was too complex. We e.g. maintained a stack of opened collections. It doesn't feel like the right way of doing stuff, tokenizer should be a simple piece of software responsible only for transforming input to the tokens. This PR brings the tokenizer to the aforementioned description.


lwronski (approved):
LGTM

PR #78, opened 2021-10-12
Parser grammar rework

kpodsiad:


yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
```diff
@@ -306,5 +306,5 @@ private[yaml] class Scanner(str: String) extends Tokenizer {
   def skipUntilNextChar() =
     while (in.isWhitespace) do in.skipCharacter()
 
-  private def skipComment(): Unit = while !in.isNewline do in.skipCharacter()
+  private def skipComment(): Unit = while (in.peek().isDefined && !in.isNewline) do in.skipCharacter()
```
lwronski, line 309: I think, that you should modify `in.isNewline` method and them check, if peek() exists.

yaml/shared/src/main/scala/org/virtuslab/yaml/Position.scala
```diff
@@ -1,9 +1,9 @@
 package org.virtuslab.yaml
 
 final case class Position(offset: Int, line: Int, column: Int, input: Vector[String]):
-  def errorMsg: String =
+  override def toString: String =
```
lwronski, line 4: Is it necessary to override the `toString` method here? Maybe, just use errorMsg method name

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
kpodsiad: It's not enough. We are skipping until we find a newline character, but for some input that will never happen. Being a new line and being defined is a different thing.

PR #79, opened 2021-10-12
Flatten test package structure

kpodsiad:


build.sbt
```diff
@@ -33,7 +34,8 @@ lazy val scalaYamlCore = crossProject(JSPlatform, JVMPlatform)
     name              := projectName,
     scalaVersion      := scala3Version,
     semanticdbVersion := scalafixSemanticdb.revision,
-    semanticdbEnabled := true
+    semanticdbEnabled := true,
+    libraryDependencies ++= Seq(Deps.pprint % Test)
   )
   .settings(docsSettings)
```
lwronski, line 40: I think, you can add here `.settings(testSettings)`

kpodsiad: `testSettings`, except dependencies, contains also `publish / skip := true` so it cannot be added.

PR #81, opened 2021-10-18
Rework tests

kpodsiad:
Simplify integration tests and their structure
Treat errors in invalid test cases as a success if obtained events are equals to expected ones.


yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
```diff
@@ -99,12 +99,14 @@ private[yaml] class Scanner(str: String) extends Tokenizer {
           in.skipN(2)
           sb.append("\"")
           readScalar()
-        case Some('"') | None =>
+        case Some('"') =>
```
lwronski, line 102: These kind of changes should be more descriptive and moved to another PR.

PR #83, opened 2021-10-19
Fix parsing implicit block key

lwronski:
Identifier of key `:` doesn't have to appear immediately after the scaler. May be separated by whitespace.


yaml/shared/src/test/scala/org/virtuslab/yaml/parser/MappingSuite.scala
```diff
@@ -351,3 +351,28 @@ class MappingSuite extends BaseParseSuite:
 
     assertEventsEquals(yaml.events, events)
   }
+
+  test("implicit block key in sequence flow") {
+    val yaml =
+      s""""implicit block key" : [
+         |  "implicit flow key" : value,
+         | ]""".stripMargin
```
kpodsiad, line 359: Align pipes with the first quote.

Issue #88, opened 2021-11-03
Handle anchors & aliases when composing nodes from events


kpodsiad:
See todo in `ComposerImpl`

Issue #87, opened 2021-11-03
Parsing anchors with keys


kpodsiad:
For
```yaml
{
  &a a : &b b,
  seq: [*a, *b]
}
```

tokenizer should return `..., MappingKey, Anchor("a"), Scalar(a), ...`, but currently it returns `..., Anchor("a"), MappingKey, Scalar(a), ...`. Returning this in the proposed order will allow us to support such constructions without any changes in the parser.

kpodsiad:
implemented in https://github.com/VirtusLab/scala-yaml/pull/91

PR #89, opened 2021-11-03
Parsing anchors & aliases

kpodsiad:
We still cannot parse anchor when a node is mapping key at the same time (require tokenizer changes from #87). Otherwise, it should just work. 

Follow-up: Handle anchors & aliases when composing nodes https://github.com/VirtusLab/scala-yaml/issues/88

### DIFF
--- a/yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
+++ b/yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
@@ -35,7 +35,9 @@ object ComposerImpl extends Composer:
         case _: Event.SequenceStart                            => composeSequenceNode(tail)
         case _: Event.MappingStart | _: Event.FlowMappingStart => composeMappingNode(tail)
         case s: Event.Scalar                                   => composeScalarNode(s, tail)
-        case event => Left(ComposerError(s"Unexpected event $event"))
+        // todo #88
+        case _: Event.Alias => Left(ComposerError(s"Aliases aren't currently supported"))
+        case event          => Left(ComposerError(s"Unexpected event $event"))
     case Nil =>
       Left(ComposerError("No events available"))
 
--- a/yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
+++ b/yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
@@ -37,7 +37,7 @@ object ComposerImpl extends Composer:
         case s: Event.Scalar                                   => composeScalarNode(s, tail)
         // todo #88
         case _: Event.Alias => Left(ComposerError(s"Aliases aren't currently supported"))
-        case event          => Left(ComposerError(s"Unexpected event $event"))
+        case event          => Left(ComposerError(s"Expected Node, but found: $event"))
     case Nil =>
       Left(ComposerError("No events available"))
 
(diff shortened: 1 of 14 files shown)


yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
```diff
@@ -35,7 +35,9 @@ object ComposerImpl extends Composer:
         case _: Event.SequenceStart                            => composeSequenceNode(tail)
         case _: Event.MappingStart | _: Event.FlowMappingStart => composeMappingNode(tail)
         case s: Event.Scalar                                   => composeScalarNode(s, tail)
-        case event => Left(ComposerError(s"Unexpected event $event"))
+        // todo #88
```
lwronski, line 38: Add more descriptive description

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
```diff
@@ -109,6 +105,7 @@ final class ParserImpl private (in: Tokenizer) extends Parser:
     def loop(events: mutable.ArrayDeque[Event]): Either[YamlError, List[Event]] = {
       getNextEvent() match
         case Right(event) =>
+          println(event)
```
lwronski, line 108: To remove

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/Event.scala
```diff
@@ -29,18 +30,25 @@ object Event:
...
+      extends Sequence
+  case class SequenceEnd(pos: Option[Position] = None) extends Sequence
 
-  sealed trait Mapping                                      extends Node
-  case class MappingStart(pos: Option[Position] = None)     extends Mapping
-  case class MappingEnd(pos: Option[Position] = None)       extends Mapping
-  case class FlowMappingStart(pos: Option[Position] = None) extends Mapping
-  case class FlowMappingEnd(pos: Option[Position] = None)   extends Mapping
+  sealed trait Mapping extends Node
+  case class MappingStart(pos: Option[Position] = None, anchor: Option[String] = None)
```
lwronski, line 49: I think, we should add Context field, which contains for example `pos` or `anchor`

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
```diff
@@ -362,21 +327,51 @@ final class ParserImpl private (in: Tokenizer) extends Parser:
...
+      nextToken.kind match
+        case TokenKind.Scalar(value, style) =>
+          in.popToken()
+          Right(Event.Scalar(value, style, Some(pos), anchor))
+        case _ =>
+          Left(ParseError.from(TokenKind.Scalar.toString, token))
+
+    inline def parseFlowNode() = parseNode(couldParseBlockColl = false)
+
+    def parseNode(couldParseBlockColl: Boolean = true): Either[YamlError, Event] =
```
lwronski, line 341:
```suggestion
    def parseNode(couldParseBlockCollection: Boolean = true): Either[YamlError, Event] =
```

lwronski (approved):
LGTM

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/Event.scala
kpodsiad: I'll do this later, in another PR.

PR #90, opened 2021-11-03
Use composition in `Event` instead of inheritance

kpodsiad:
There might be a side effect of this PR: complex keys may work from now, but I haven't testes this yet.


lwronski (approved):
LGTM

PR #91, opened 2021-11-07
Fix parsing anchor in mapping block

lwronski:


yaml/shared/src/test/scala/org/virtuslab/yaml/parser/AnchorSpec.scala
```diff
@@ -158,7 +141,19 @@ class AnchorSpec extends BaseYamlSuite:
...
+      DocumentStart(explicit = true),
+      MappingStart(),
+      Scalar("hr"),
+      SequenceStart(),
+      Scalar("Mark McGwire"),
+      Scalar("Sammy Sosa", metadata = NodeEventMetadata(Anchor("SS"))),
+      SequenceEnd,
+      Scalar("rbi"),
+      SequenceStart(),
+      Alias(Anchor("SS")),
```
kpodsiad, line 153: We may want to change the anchor name from `SS` to something else. It's better to avoid such abbreviations, even when example comes from the yaml website

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/ReaderCtx.scala
```diff
@@ -18,6 +18,7 @@ case class ReaderCtx(reader: Reader) {
 
   def indent: Int                     = indentations.lastOption.getOrElse(-1)
   def addIndent(newIndent: Int): Unit = indentations.append(newIndent)
+  def removeLastIndent(): Unit        = indentations.removeLast()
```
kpodsiad, line 21: Wouldn't `removeLastOption` be safer?

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
```diff
@@ -109,9 +108,22 @@ private[yaml] class Scanner(str: String) extends Tokenizer {
...
+  private def parseAnchor(): List[Token] =
+    val (name, anchorPos) = parseAnchorName()
+    val nexTokens         = getNextTokens()
+
+    val anchorToken = Token(Anchor(name), anchorPos)
+    nexTokens match {
+      case (Token(_: MappingStart.type, _) |
+          Token(_: FlowMappingStart.type, _)) :: Token(_: MappingKey.type, _) :: rest =>
+        ctx.removeLastIndent()
+        ctx.addIndent(anchorPos.column)
```
kpodsiad, line 120: So here we have to adjust the indentation level because that added by the mapping isn't the right one, correct?

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
```diff
@@ -109,9 +108,22 @@ private[yaml] class Scanner(str: String) extends Tokenizer {
...
+        nexTokens.take(2) ::: anchorToken +: rest
+      case Token(_: MappingKey.type, _) :: rest
+          if (ctx.indent == anchorPos.column || ctx.isInFlowCollection) =>
+        nexTokens.take(1) ::: anchorToken +: rest
```
kpodsiad, line 124: Which situation is handled by this case?

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
```diff
@@ -305,33 +317,33 @@ private[yaml] class Scanner(str: String) extends Tokenizer {
...
-          )
+        val maybeMappingStart =
+          if (ctx.indent < scalar.pos.column && !ctx.isInFlowCollection) then
+            ctx.addIndent(scalar.pos.column)
+            List(Token(MappingStart, scalar.pos))
+          else Nil
+
+        maybeMappingStart :+ Token(MappingKey, scalar.pos) :+ scalar :+ Token(
+          MappingValue,
+          scalar.pos
```
kpodsiad, line 344:
```suggestion
        maybeMappingStart ++ List(
          Token(MappingKey, scalar.pos),
          scalar,
          Token(
            MappingValue,
            scalar.pos
          )
        )
```

kpodsiad (approved):
LGTM

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/ReaderCtx.scala
```diff
@@ -18,6 +18,7 @@ case class ReaderCtx(reader: Reader) {
...
 
   def checkIndents(current: Int): Unit =
```
kpodsiad, line 23: `checkIndents` could return `List[Token]` instead of appending those token itself

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
lwronski: Yes, you are right. The anchor position indicates the mapping indent.

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
lwronski:
```
key: value
&a a: &b b
```

PR #93, opened 2021-11-09
Validate key in mapping - not allow to multine in simple key

lwronski:


yaml/shared/src/main/scala/org/virtuslab/yaml/Position.scala
```diff
@@ -1,11 +1,24 @@
 package org.virtuslab.yaml
 
-final case class Position(offset: Int, line: Int, column: Int, input: Vector[String]):
+final case class Position(
+    offset: Int,
+    line: Int,
+    column: Int,
+    input: Vector[String],
+    endLine: Option[Int] = None,
+    endColumn: Option[Int] = None
```
kpodsiad, line 9: Hmm, I've got a problem with adding `endline` and `endColumn`. When I think about it, now `Position` becomes effectively a `Range` with an optional end Position parameter. Thus, maybe following 
```scala
final case class Position(
    offset: Int,
    line: Int,
    column: Int,
)

final case class Range(
    input: Vector[String],
    start: Position,
    end: Option[Position]
)
```
would be a better model for this?

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
```diff
@@ -4,35 +4,38 @@ import org.virtuslab.yaml.internal.load.reader.token.{BlockChompingIndicator, Sc
...
+      ctx.tokens.head
+    }
 
-  private def getNextTokens(): List[Token] =
+  private def getNextTokens(): Either[YamlError, List[Token]] =
     skipUntilNextToken()
     val closedTokens = ctx.checkIndents(in.column)
     val peeked       = in.peek()
-    val tokens = peeked match
+    val tokens: Either[YamlError, List[Token]] = peeked match
```
kpodsiad, line 38: Crazy idea, wdyt about throwing an exception without stack trace and cathing it in `getToken()`? It would reduce the amount of `Right(...)` needed to be written and probably save a few allocations.

kpodsiad (commented):
It looks promising, but there are a few places that need more discussion.

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Tokenizer.scala
kpodsiad: or if we want to avoid exceptions, I think that union type `YamlError | List[Token]` is also a promising option.

yaml/shared/src/main/scala/org/virtuslab/yaml/Range.scala
```diff
@@ -0,0 +1,28 @@
...
+    endColumn: Option[Int] = None
+):
+  def errorMsg: String =
+    val msg          = input(start.line)
+    val spaces       = start.column
+    val circumflexes = msg.length - spaces
+    s"""|$msg
+        |${" " * spaces}^""".stripMargin
+
+  def addEndPos(end: Range): Range =
```
kpodsiad, line 22:
```suggestion
  def withEndPos(end: Range): Range =
```

yaml/shared/src/main/scala/org/virtuslab/yaml/Range.scala
```diff
@@ -0,0 +1,28 @@
...
+    offset: Int,
+    line: Int,
+    column: Int
+)
+
+final case class Range(
+    start: Position,
+    input: Vector[String],
+    endLine: Option[Int] = None,
+    endColumn: Option[Int] = None
```
kpodsiad, line 13: wdyt about?
```suggestion
    start: Position,
    end: Option[Position],
    input: Vector[String],
```

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Reader.scala
```diff
@@ -21,15 +21,15 @@ trait Reader:
   def line: Int
   def column: Int
   def offset: Int
-  def pos: Position
+  def pos: Range
```
kpodsiad, line 24:
```suggestion
  def pos: Position
  def range(start: Position = in.pos, end: Position = in.pos): Range
```
thus will allow:
```scala
val start = in.pos
// parse something
val range = in.range(start = start)
```
wdyt?

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/token/Token.scala
```diff
@@ -1,9 +1,9 @@
 package org.virtuslab.yaml.internal.load.reader.token
 
-import org.virtuslab.yaml.Position
+import org.virtuslab.yaml.Range
 import org.virtuslab.yaml.internal.load.reader.token.ScalarStyle
 
-final case class Token(kind: TokenKind, pos: Position)
+final case class Token(kind: TokenKind, pos: Range)
```
kpodsiad, line 6:
```suggestion
final case class Token(kind: TokenKind, range: Range)
```

kpodsiad (approved):
Looks awesome and CI is green. LGTM!

PR #94, opened 2021-11-10
Scalafix

kpodsiad:


.scalafix.conf
```diff
@@ -0,0 +1,14 @@
+rules = [
```
lwronski, line 1:
```suggestion
rules = [ 
  DisableSyntax,
  RemoveUnused,
  OrganizeImports,
  NoValInForComprehension,
  ProcedureSyntax
]

DisableSyntax.noFinalize = true
DisableSyntax.noReturns = true



```

lwronski (approved):
LGTM, I'm only propose to add the other rules.

PR #98, opened 2021-11-13
Add tags to nodes

kpodsiad:
Each YAML node except its kind (scalar, mapping, sequence) has also a tag. Tag property indicates the type that the application should use to represent a given node. Typically, most tags are not explicitly specified in the character stream so YAML processor has to resolve them on its own using some schema. Recommended done is [core schema](https://yaml.org/spec/1.2.2/#core-schema).

- [ ] Add more testcases (e.g. all form [this chapter](https://yaml.org/spec/1.2.2/#chapter-10-recommended-schemas))
- [ ] `YamlDecoder` should be easier to create. It'll be good to have some factory functions which will reduce required boilerplate (maybe as a follow-up PR?)

### DIFF
--- a/yaml/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
+++ b/yaml/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
@@ -1,0 +1,42 @@
+package org.virtuslab.yaml
+
+import scala.reflect.ClassTag
+
+final case class Tag(value: String)
+
+object Tag:
+  def apply[T](implicit classTag: ClassTag[T]): Tag = Tag(s"!${classTag.runtimeClass.getName}")
+
+  private val default = "tag:yaml.org,2002:"
+  val nullTag: Tag    = Tag(s"${default}null")
+  val boolean: Tag    = Tag(s"${default}bool")
+  val int: Tag        = Tag(s"${default}int")
+  val float: Tag      = Tag(s"${default}float")
+  val str: Tag        = Tag(s"${default}str")
+  val seq: Tag        = Tag(s"${default}seq")
+  val map: Tag        = Tag(s"${default}map")
+
+  private val nullPattern    = "null|Null|NULL|~".r
+  private val booleanPattern = "true|True|TRUE|false|False|FALSE".r
+  private val int10Pattern   = "[-+]?[0-9]+".r
+  private val int8Pattern    = "0o[0-7]+".r
+  private val int16Pattern   = "0x[0-9a-fA-F]+".r
+  private val floatPattern   = "[-+]?(\\.[0-9]+|[0-9]+(\\.[0-9]*)?)([eE][-+]?[0-9]+)?".r
+  private val minusInfinity  = "-(\\.inf|\\.Inf|\\.INF)".r
+  private val plusInfinity   = "\\+?(\\.inf|\\.Inf|\\.INF)".r
+
+  def resolveTag(value: String): Tag = {
+    value match {
+      case null               => nullTag
+      case nullPattern(_*)    => nullTag
+      case booleanPattern(_*) => boolean
+      case int10Pattern(_*)   => int
+      case int8Pattern(_*)    => int
+      case int16Pattern(_*)   => int
+      case floatPattern(_*)   => float
+      case minusInfinity(_*)  => float
+      case plusInfinity(_*)   => float
+      case _                  => str
+    }
+  }
+end Tag
--- a/yaml/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
+++ b/yaml/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
@@ -16,6 +16,8 @@ object Tag:
   val seq: Tag        = Tag(s"${default}seq")
   val map: Tag        = Tag(s"${default}map")
 
+  val primitives = Set(nullTag, boolean, int, float, str)
+
   private val nullPattern    = "null|Null|NULL|~".r
   private val booleanPattern = "true|True|TRUE|false|False|FALSE".r
   private val int10Pattern   = "[-+]?[0-9]+".r
--- a/yaml/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
+++ b/yaml/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
@@ -2,21 +2,28 @@ package org.virtuslab.yaml
 
 import scala.reflect.ClassTag
 
-final case class Tag(value: String)
+sealed trait Tag:
+  def value: String
+
+final case class CoreSchemaTag(value: String) extends Tag
+final case class CustomTag(value: String)     extends Tag
 
 object Tag:
-  def apply[T](implicit classTag: ClassTag[T]): Tag = Tag(s"!${classTag.runtimeClass.getName}")
+  def apply[T](implicit classTag: ClassTag[T]): Tag = CustomTag(
+    s"!${classTag.runtimeClass.getName}"
+  )
 
   private val default = "tag:yaml.org,2002:"
-  val nullTag: Tag    = Tag(s"${default}null")
-  val boolean: Tag    = Tag(s"${default}bool")
-  val int: Tag        = Tag(s"${default}int")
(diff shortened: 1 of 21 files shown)


yaml/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
```diff
@@ -12,64 +12,96 @@ import org.virtuslab.yaml.Node.*
...
+        case MappingNode(mappings, Tag.map) =>
+          val decoder = summon[YamlDecoder[Map[Any, Any]]]
+          decoder.construct(node)
+        case SequenceNode(seq, Tag.seq) =>
+          val decoder = summon[YamlDecoder[Seq[Any]]]
+          decoder.construct(node)
+        case _ =>
+          settings.constructors.get(node.tag) match {
+            case Some(decoder) => decoder.construct(node)
+            case None => Left(ConstructError(s"Could't create construct instance for ${node.tag}"))
```
lwronski, line 95: This message should contains more helpful text, for example that user should create own instance YamlDecoder for Tag.

yaml/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
```diff
@@ -0,0 +1,44 @@
...
+  private val default = "tag:yaml.org,2002:"
+  val nullTag: Tag    = Tag(s"${default}null")
+  val boolean: Tag    = Tag(s"${default}bool")
+  val int: Tag        = Tag(s"${default}int")
+  val float: Tag      = Tag(s"${default}float")
+  val str: Tag        = Tag(s"${default}str")
+  val seq: Tag        = Tag(s"${default}seq")
+  val map: Tag        = Tag(s"${default}map")
+
+  val primitives = Set(nullTag, boolean, int, float, str)
```
lwronski, line 19: wdyt about creating special type for primitives, 

```
trait Tag:
   def: value 
   
final case class PrimitiveTag(value: String)
final case class CustomTag(value: String)

```

yaml/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
```diff
@@ -12,64 +12,96 @@ import org.virtuslab.yaml.Node.*
...
+      }
+    }
   }
 
   given [T](using c: YamlDecoder[T]): YamlDecoder[Option[T]] = YamlDecoder {
-    case ScalarNode(value, _) =>
+    case ScalarNode(value, tag) =>
       value match
-        case "null" | "" => Right(None)
+        case _ if tag == Tag.nullTag => Right(None)
```
lwronski, line 104: I don't think, we should remove `case "null" | ""`, we should allow to specify `null` with "null" or ""

yaml/shared/src/test/scala/org/virtuslab/yaml/decoder/DecoderSuite.scala
```diff
@@ -184,3 +185,49 @@ class PrimitiveDecoderSuite extends munit.FunSuite:
...
+    val expected = Custom(5, 10)
+
+    val decoder = new YamlDecoder[Custom] {
+      override def construct(node: Node)(using
+          settings: LoadSettings = LoadSettings.empty
+      ): Either[ConstructError, Custom] = node match {
+        case ScalarNode(value, _) =>
+          val int = value.toInt
+          Right(Custom(int, int * 2))
+        case _ => ???
```
lwronski, line 224: It shouldn't be necessary. `Partial Functions` can be helpful in this case.

yaml/shared/src/test/scala/org/virtuslab/yaml/decoder/DecoderSuite.scala
```diff
@@ -184,3 +185,49 @@ class PrimitiveDecoderSuite extends munit.FunSuite:
...
+      123      -> 321,
+      "string" -> "aezakmi",
+      true     -> false,
+      5.5      -> 55.55
+    )
+
+    assertEquals(yaml.as[Map[Any, Any]], Right(expected))
+  }
+
+  test("decode using custom tag".only) {
```
lwronski, line 208: `Only` is not allowed in CI.

yaml/shared/src/test/scala/org/virtuslab/yaml/ConstructSuite.scala
```diff
@@ -45,3 +45,19 @@ class ConstructSuite extends munit.FunSuite:
...
+
+    val node = MappingNode(
+      SequenceNode(1, 2) -> ScalarNode("seq")
+    )
+
+    val expected = Map[Any, Any](
+      Seq(1, 2) -> "seq"
+    )
+
+    println(SequenceNode(1, 2))
```
lwronski, line 59: To remove.

yaml/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
```diff
@@ -86,14 +88,16 @@ object ComposerImpl extends Composer:
...
-            case Left(err)         => Left(err)
+            case Right(node @ (key, value), rest) => parseMappings(rest, mappings :+ node, key.pos)
+            case Left(err)                        => Left(err)
     }
 
     parseMappings(events, Nil).map { case (Result(nodes, rest), pos) =>
-      Result(Node.MappingNode(nodes, pos), rest)
+      Result(
+        new Node.MappingNode(nodes.toMap, Tag.map, pos),
+        rest
```
lwronski, line 100: `new` is necessary in this case?

yaml/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
```diff
@@ -12,64 +12,96 @@ import org.virtuslab.yaml.Node.*
...
+            case Tag.boolean => value.toBooleanOption.toRight(cannotParse(value, "Boolean", node))
+            case Tag.int =>
+              if value.startsWith("0b") then
+                Try(Integer.parseInt(value.drop(2), 8)).toEither.swap
+                  .map(t => ConstructError.from(t, "Int", node))
+                  .swap
+              else if value.startsWith("0x") then
+                Try(Integer.parseInt(value.drop(2), 8)).toEither.swap
+                  .map(t => ConstructError.from(t, "Int", node))
+                  .swap
```
lwronski, line 80: The same above:

```suggestion
.left.map(t => ConstructError.from(t, "Int", node))
```

lwronski (approved):

yaml/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
kpodsiad: but `""` is an empty string, not a `null` value, we don't differentiate this but we should. Maybe in `ParserImpl#parseNode` we should alter the last case to always emit `null` tag with an empty string? Thanks to that we would be able to recognize explicit empty string `("", tag:yaml.org,2002:str)` (because we parsed Scalar token and the tag will resolve to string) from `("", tag:yaml.org,2002:null)` (because we expect one of the node's tokens, but we didn't encounter any, so we will explicitly add null tag)

lwronski: It's make sense, I approve.

yaml/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
```diff
@@ -64,20 +69,19 @@ object YamlDecoder:
 
   given YamlDecoder[Any] = new YamlDecoder {
     def construct(node: Node)(using settings: LoadSettings = LoadSettings.empty) = {
+      // println(settings)
       node match {
-        case ScalarNode(value, tag) if Tag.primitives.contains(tag) =>
+        case ScalarNode(value, tag: PrimitiveTag) if Tag.primitives.contains(tag) =>
```
lwronski, line 74: `Tag.primitives.contains(tag)` I don't think, it is required,.

lwronski (approved):
LGTM

PR #99, opened 2021-11-16
Add API for modify Node

lwronski:
Expose API for modify nod, for example:
   `node.modify("services")("web")("ports")(0).setValue("6000:6000")`


yaml/shared/src/test/scala/org/virtuslab/yaml/TestOps.scala
```diff
@@ -0,0 +1,14 @@
...
+object TestOps {
+
+  implicit class EitherThrowOps[E <: YamlError, T](private val either: Either[E, T])
+      extends AnyVal {
+    def orThrow: T =
+      either match {
+        case Left(e)  => throw new RuntimeException(e.msg)
+        case Right(t) => t
+      }
+  }
```
kpodsiad, line 12: Could be written as a extension method

yaml/shared/src/main/scala/org/virtuslab/yaml/NodeVisitor.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import org.virtuslab.yaml.Node
+import org.virtuslab.yaml.Node.*
+import org.virtuslab.yaml.Range
+import org.virtuslab.yaml.TraverseError
+import org.virtuslab.yaml.YamlError
+import org.virtuslab.yaml.syntax.NodeSelector
+import org.virtuslab.yaml.syntax.NodeSelector.*
+
+case class NodeVisitor(node: Node, selectors: List[NodeSelector]):
```
kpodsiad, line 11:
```suggestion
class NodeVisitor private (node: Node, selectors: List[NodeSelector]):
```
+ `def apply(node: Node, selectors: List[NodeSelector]): NodeVisitor = new NodeVisitor(node, selectors)` in companion object
There is no reason for `NodeVisitor` to be a case class. Moreover, it exposes copy method.

kpodsiad (approved):

PR #100, opened 2021-11-16
Aliases

kpodsiad:
Aliases are now took into account when composing data structures from nodes


lwronski (approved):
LGTM, I think this case is worth testing:
```
- &a
  name: Mark McGwire
  hr:   65
- *a
```

PR #126, opened 2022-01-06
Remove node using modify API

lwronski:


core/shared/src/main/scala/org/virtuslab/yaml/NodeVisitor.scala
```diff
@@ -15,14 +15,30 @@ class NodeVisitor(node: Node, selectors: List[NodeSelector]):
...
+          )
+        )
+    }
+
+  private def removeScalarNode(scalar: ScalarNode): Either[TraverseError, ScalarNode] =
+    selectors match {
+      case Nil =>
+        Right(
+          scalar.copy(
+            value = ""
```
kpodsiad, line 41: maybe it should tag to null?

core/shared/src/main/scala/org/virtuslab/yaml/NodeVisitor.scala
```diff
@@ -15,14 +15,30 @@ class NodeVisitor(node: Node, selectors: List[NodeSelector]):
...
       case Nil =>
         Right(
           scalar.copy(
-            value = value
+            value = modifyValue(scalar.value)
+          )
+        )
+      case _ =>
+        Left(
+          TraverseError(
```
kpodsiad, line 30: What do you think about renaming `TraverseError` into `ModifyError`?

core/shared/src/main/scala/org/virtuslab/yaml/NodeVisitor.scala
```diff
@@ -52,7 +68,32 @@ class NodeVisitor(node: Node, selectors: List[NodeSelector]):
...
+          )
+        } else {
+          val nodes      = sequence.nodes
+          val updateNode = NodeVisitor(nodes(index), rest).removeValue()
+          updateNode.map(node => sequence.copy(nodes = nodes.updated(index, node)))
+        }
+      case StringSelector(field) :: rest =>
+        Left(
+          TraverseError(
+            s"Expeceted index of sequence, insted found string path: ${field}"
```
kpodsiad, line 90: `Expeceted` -> `Expected`, `insted` -> `instead`
Also, maybe following? 
```suggestion
            s"Found string path: ${field} traversing sequence, but index was expected"
```

kpodsiad (approved):
We'll continue working on this in following PR's

Issue #159, opened 2022-07-04
Publish for 2.12/2.13?


armanbilge:
Thrilled to discover this project yesterday! It scratches a real itch :)

I understand the codec derivation features rely on Scala 3, but I assume it should be straightforward to cross-compile the parser and `Node` AST for Scala 2.12/2.13? The parser and `Node` AST is sufficient to enable a [circe-yaml](https://github.com/circe/circe-yaml) frontend for example.

Happy to do the work to split into version-specific sources if 👍 

Discussion on the importance of 2.12 (and 2.13) support is in:
- https://github.com/circe/circe/pull/1818

I noticed that the advice in https://github.com/VirtusLab/scala-yaml/issues/102#issuecomment-986246214 is to use `CrossVersion.for2_13Use3` but the consensus seems to be to avoid this technique for libraries. See also:
- https://gist.github.com/djspiewak/b8f2b4547951442102488964bc351cf9#5-dont-rely-on-withdottycompat
- https://github.com/scala-js/scala-js-dom/issues/451

Thanks!

armanbilge:
Yeah, AFAICT `for2_13Use3` just doesn't work. Example:
```scala
//> using scala "2.13.8"
//> using lib "org.virtuslab:scala-yaml_3:0.0.4"
//> using option "-Ytasty-reader"

object Main {
  val myYaml = """foo: - bar"""
  def main(args: Array[String]): Unit =
    org.virtuslab.yaml.asNode(myYaml)
}
```
```
Compiling project (Scala 2.13.8, JVM)
[error] ./bug.scala:8:5: object asNode is not a member of package org.virtuslab.yaml
[error] did you mean Node?
[error]     org.virtuslab.yaml.asNode(myYaml)
[error]     ^^^^^^^^^^^^^^^^^^^^^^^^^
Error compiling project (Scala 2.13.8, JVM)
```

kpodsiad:
Hey @armanbilge, feel free to do it. Scala 2 is still the largest piece of cake in ecosystem and enabling to work with YAML in ScalaJS and ScalaNative might be useful for some people. Piggybacking on circe encoders should cover almost all of use cases.

This was more a fun & research project, but I don't see a reason why not shape it into more mature and stable one.

armanbilge:
@kpodsiad thanks! I opened https://github.com/VirtusLab/scala-yaml/pull/162.

> This was more a fun & research project, but I don't see a reason why not shape it into more mature and stable one.

[...]

PR #162, opened 2022-07-07
Cross-build for Scala 2.13

armanbilge:
Closes https://github.com/VirtusLab/scala-yaml/issues/159.

I had a brief look at adding 2.12 as well, but that still requires more changes. 2.13 is pretty good for now :)


kpodsiad:
Seems like this PR is on good way @armanbilge ;)

build.sbt
```diff
@@ -36,13 +37,11 @@ inThisBuild(
 
 ThisBuild / scalafixDependencies += "com.github.liancheng" %% "organize-imports" % "0.6.0"
 
-lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+lazy val core = crossProject(JSPlatform, JVMPlatform /*, NativePlatform*/ )
```
kpodsiad, line 40: Is there a reason for commenting native version?

kpodsiad (commented):
scalafmt and native fails on CI, after they're green I think this will be good to go 👍

build.sbt
armanbilge: Yes, because I had to bump to Scala 3.2.0-RC2 the Native plugin is not yet published for these versions.

kpodsiad: CI for for native can be disabled (for a while) to not block this PR

armanbilge: @kpodsiad cool, thanks, I've fixed formatting and disabled Native CI for now.

kpodsiad (approved):
Looks good, CI is green 👍

Issue #177, opened 2022-09-24
jar for 0.0.5 version for scala 2.13 is empty 


Morgaroth:
please see jar file published for 2.13:0.0.5 is empty (weights 300B - only META-INF dir)

https://repo1.maven.org/maven2/org/virtuslab/scala-yaml_2.13/0.0.5/scala-yaml_2.13-0.0.5.jar

gauravkumar37:
+ @kpodsiad 
confirmed, please fix

armanbilge:
Thanks for the ping, fix in https://github.com/VirtusLab/scala-yaml/pull/178.

PR #178, opened 2022-10-07
No-publish the root

armanbilge:
Pretty sure this fixes https://github.com/VirtusLab/scala-yaml/issues/177. The (empty) root project was being published with the same name as the core module, causing it to overwrite with an empty jar.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,6 +6,8 @@ def projectName          = "scala-yaml"
 def localSnapshotVersion = "0.0.6-SNAPSHOT"
 def isCI                 = System.getenv("CI") != null
 
+enablePlugins(NoPublishPlugin)
+
 inThisBuild(
   List(
     organization       := "org.virtuslab",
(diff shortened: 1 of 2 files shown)


gauravkumar37:
Thanks for looking into this. Unfortunately, I still see empty jars.
I have checked out your PR, and ran `sbt clean +package`. The jars present in `target/scala*` directories are still 318 bytes only.

armanbilge:
I can't speak to your results. I ran `publishLocal` and it looks fine to me.

armanbilge:
@gauravkumar37 are you looking in `target/scala*` or `core/jvm/target/scala`? The former is supposed to be empty, the latter is not.

armanbilge:
Right, the problem was that the empty root `target/` was overriding the `core/jvm/target`. No-publishing the root fixes that.

gauravkumar37:
Thanks.

Off topic to this PR, I couldn't find any docs for scala 2.13 usage. The official guide is for scala 3 syntax.

armanbilge:
Scala 2.13 was a later addition in https://github.com/VirtusLab/scala-yaml/pull/162, this project was originally built targeting Scala 3. I am sure a docs PR would be welcome :)

kpodsiad:
Thank you @armanbilge for tackling this, I couldn't find time to look at this myself.

PR #195, opened 2022-11-13
small tweaks in tokenizer

kpodsiad:
Main change - return [null character](https://www.ascii-code.com/0) instead of None in reader and tokenizer. I'm planning to make tokenizer less magical (well, other parts of code too) and I found wrapping with Some more difficult to comprehend than just operating on plain chars

### DIFF
--- a/core/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Reader.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Reader.scala
@@ -57,7 +57,7 @@ private[yaml] class StringReader(in: String) extends Reader {
 
   override def peek(n: Int = 0): Char =
     if (offset + n < in.length) in.charAt(offset + n)
-    else '\u0000'
+    else Reader. nullTerminator
 
   private def nextLine() = { column = 0; line += 1 }
   private def skipAndMantainPosition() = {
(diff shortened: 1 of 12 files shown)


core/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/Reader.scala
```diff
@@ -35,18 +55,9 @@ private[yaml] class StringReader(in: String) extends Reader {
...
-  override def peekN(n: Int): String    = (0 until n).map(peek(_)).flatten.mkString("")
-  override def isWhitespace             = peek().exists(_.isWhitespace)
-  override def isNextWhitespace         = peekNext().exists(_.isWhitespace)
-  override def isNewline                = peek().exists(c => c == '\n' || isWindowsNewline(c))
-  override def isNextNewline            = peekNext().exists(c => c == '\n' || isWindowsNewline(c))
-
-  private def isWindowsNewline(c: Char) = c == '\r' && peekNext().exists(_ == '\n')
+  override def peek(n: Int = 0): Char =
+    if (offset + n < in.length) in.charAt(offset + n)
+    else '\u0000'
```
lwronski, line 60:
```suggestion
    else Reader. nullTerminator
```

lwronski (approved):

Issue #220, opened 2023-04-25
String types should not be silently converted to `Integer` or `Boolean`


winitzki:
If a Yaml value is a `String` but the actual value appears to be parsable as integer or boolean, `scala-yaml` automatically translates it into `java.lang.Integer` and `java.lang.Boolean`. This changes the yaml schema silently and unexpectedly. With typical kubernetes Yaml where values such as `"2"` or `"true"` are of type `String` , the Yaml cannot be parsed after `scala-yaml` serializes the data.

Example code:

```scala
    val yaml1 = """  revision: "2"  """.as[Any].toOption.get
   // We expect `revision` to be of type `String` and value `"2"`, instead have an integer:
    assert(yaml1.asInstanceOf[Map[String, Any]].apply("revision").isInstanceOf[java.lang.Integer])

    val yaml2 = """  enabled: "true"  """.as[Any].toOption.get
   // We expect `enabled` to be of type `String` and value `"true"`, instead have a Boolean:
    assert(yaml2.asInstanceOf[Map[String, Any]].apply("enabled").isInstanceOf[java.lang.Boolean])
```

I expected it to be a `String` in each case, since `"2"` and `"true"` are just strings.

Not sure if this is a feature that can be turned off, please advise.

PR #221, opened 2023-04-27
Issue #220 - do not convert quoted strings into other primitive types

winitzki:
Yaml values of the form `"bool"` or `"123"` or other quoted strings will not be interpreted as `java.lang.Boolean` or `java.lang.Integer`.

This fixes Issue #220


lwronski (approved):
Thanks @winitzki LGTM

PR #234, opened 2023-06-27
Parses a nonexistent key as None

### LINKED ISSUE #203: Don't require `!!null`  for `Option[T]`
I see there is some work to add a `YamlEncoder` for `Option` in https://github.com/VirtusLab/scala-yaml/pull/127/files but even in that situation when using an `Option[T]` you still have to use `!!null` inside your yml file which is clunky.

For example if you have something like this:

```scala
//> using scala "3.2.1"
//> using lib "org.virtuslab::scala-yaml:0.0.6"

import scala.io.Source
import Encoders.given_YamlDecoder_Option
import Encoders.given_YamlEncoder_Option
import org.virtuslab.yaml.YamlCodec
import org.virtuslab.yaml.YamlEncoder
import org.virtuslab.yaml.Node
import org.virtuslab.yaml.YamlDecoder
import org.virtuslab.yaml.Node.ScalarNode
import org.virtuslab.yaml.*

@main def run() =

  val names = Source.fromFile("./example.yml").mkString
  val parsed = names.as[Seq[Name]]
  parsed match
    case Left(e) =>
      println(e.msg)
    case Right(value) =>
      print(s"got ${value} names")

final case class Name(
    name: String,
    other: Option[String]
) derives YamlCodec

object Encoders:
  given [T](using encoder: YamlEncoder[T]): YamlEncoder[Option[T]] = {
    case Some(t) => encoder.asNode(t)
    case None    => Node.ScalarNode("!!null")
  }

  // I know this is weird, but just trying to see if there is any way to get this to be correct if the key doesn't exist.
  given [T](using decoder: YamlDecoder[T]): YamlDecoder[Option[T]] =
    YamlDecoder { node => Right(None) }

```

And you have a file like this:

```yml
- name: My name
  other: My other name

- name: Their name
```

It will fail with:

```
❯ scli run .
Key other doesn't exist in parsed document
```

Unless I'm doing something wrong, I wish the missing key could just be decoded into a `None`. Is there a way to accomplish this with a custom decoder?

andrelfpinto:
Implicitly parses a nonexistent key as None in case of key type is Option[_]
Fix #203

### DIFF
--- a/core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
+++ b/core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
@@ -48,8 +48,8 @@ private[yaml] trait DecoderMacros {
   }
 
   protected inline def deriveProduct[T](p: Mirror.ProductOf[T]) =
-    val instances  = summonAll[p.MirroredElemTypes]
-    val elemLabels = getElemLabels[p.MirroredElemLabels]
+    val instances     = summonAll[p.MirroredElemTypes]
+    val elemLabels    = getElemLabels[p.MirroredElemLabels]
     val optionalTypes = getOptionalTypes[p.MirroredElemTypes]
     new YamlDecoder[T] {
       override def construct(node: Node)(using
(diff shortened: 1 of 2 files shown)


lwronski (approved):
@andrelfpinto Thanks for your contribution. LGTM.

PR #271, opened 2024-03-13
Improve `YamlError` ADT

sherpal:
I am opening this pull request to discuss the idea of improving (and "opening") the `YamlError` ADT.

Currently, only "ascii" messages are provided to the end users. So after calling a method like `asNode` on a `String`, all underlying information (position of the error, token responsible, expected token kind) are no more available.

The `ParseError` is the most obvious enhancement, but the same philosophy can probably be applied to others, like the `ComposeError`, for example.

This will be a (slight) breaking change (in the case people were destructuring the `ParseError` case class in a match statement, for example) but since it will only lead to compile-time error, I don't think it's a big deal.

I'm adding a regression test to verify that the actual error message did not change (and will not, at least not unwillingly).

### DIFF
--- a/core/shared/src/main/scala/org/virtuslab/yaml/YamlError.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/YamlError.scala
@@ -5,6 +5,7 @@ import scala.util.control.NoStackTrace
 
 import org.virtuslab.yaml.internal.load.reader.token.Token
 import org.virtuslab.yaml.internal.load.reader.token.TokenKind
+import org.virtuslab.yaml.internal.load.TagValue
 
 /**
  * An ADT representing a decoding failure.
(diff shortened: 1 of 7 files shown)


roman-mibex-2:
Also having it on the 'ConstructError' would help. When you construct data types from the YAML-Nodes and fail, you want to know where you failed.

sherpal:
I added the same treatment to `ConstructError` and `ScannerError`, though I feel less confident for the design in the `ConstructError` case.

lbialy:
better question here is: are these all of the changes that users could possibly have for errors? it would be probably good to do this change once since it breaks b/compat

kpodsiad:
@tgodzik looks fine :)

sherpal:
> better question here is: are these all of the changes that users could possibly have for errors? it would be probably good to do this change once since it breaks b/compat

It would be nice, indeed, Do you see any other things were we should/could add more information?

lbialy:
Is there a reason why you'd not put the failing Node instance to `ConstructError.from`? There's a variant that takes `Node` as a second argument and it would probably yield position information for nodes that do have that (or maybe I'm missing something here).

core/shared/src/main/scala/org/virtuslab/yaml/YamlError.scala
```diff
@@ -13,39 +14,49 @@ sealed trait YamlError {
...
+  }
+
+  final case class NoRegisteredTagDirective(handleKey: String, tokenTag: Token) extends ParseError {
+    def msg: String = s"There is no registered tag directive for handle $handleKey"
+  }
 }
 
 final case class ComposerError(msg: String) extends YamlError
 
 final case class ModifyError(msg: String) extends YamlError
```
kpodsiad, line 36: for future: ModifyError is shouldn't rather be part of `YamlError` hierarchy. cc: @lbialy

lbialy: Could you quickly enumerate what each of these error types is meant to represent and when can they occur?

lbialy:
> Is there a reason why you'd not put the failing Node instance to `ConstructError.from`? There's a variant that takes `Node` as a second argument and it would probably yield position information for nodes that do have that (or maybe I'm missing something here).

@sherpal

sherpal:
> Is there a reason why you'd not put the failing Node instance to `ConstructError.from`? There's a variant that takes `Node` as a second argument and it would probably yield position information for nodes that do have that (or maybe I'm missing something here).

(Apologies, I was on vacation and I missed the notif!)

Do you mean something like [here](https://github.com/VirtusLab/scala-yaml/pull/271/files#diff-32973d256749bcd5d6f523448e86986a9ce9483feb6047a712ae41b7e23d2670R43)? I would love to, but I am not sure the information is available there. Or am I mistaken?

lbialy:
More like https://github.com/VirtusLab/scala-yaml/pull/271/files#diff-32973d256749bcd5d6f523448e86986a9ce9483feb6047a712ae41b7e23d2670R22 where it can be 
`case node => Left(ConstructError.from(s"Parameter of a class must be a scalar value", node))` to preserve positioning in errors. Don't worry, I'll handle this along with errors.

lbialy:
Superseded by #309

Issue #119, opened 2021-12-30
Add `YamlEncoder` for `Option`


kpodsiad:
```scala
// using scala 3.1.0
// using lib org.virtuslab::scala-yaml:0.0.4

import org.virtuslab.yaml.*

case class Foo(field: Option[String]) derives YamlCodec
```
running via `scala-cli run .` yields compilation error
```
Compiling project (Scala 3.1.0, JVM)
[error] ./example2.sc:6:47: no implicit argument of type org.virtuslab.yaml.YamlEncoder[Option[String]] was found
[error] case class Foo(field: Option[String]) derives YamlCodec
[error]                                               ^
Error compiling project (Scala 3.1.0, JVM)
```

dubinsky:
Hi @kpodsiad! Do you think there is any hope for this bug to ever get fixed? Thanks!

lbialy:
@dubinsky I'll take a look at all the PRs and issues this week.

dubinsky:
Thank you @lbialy!

PR #296, opened 2024-05-10
Option encoder

lbialy:
Supersedes #127 (thanks @kpodsiad), closes https://github.com/VirtusLab/scala-yaml/issues/119


core/shared/src/test/scala-3/org/virtuslab/yaml/BaseYamlEncoderSuite.scala
```diff
@@ -125,6 +125,16 @@ class BaseYamlEncoderSuite extends munit.FunSuite:
     assertEquals(data.asYaml, expected)
   }
 
+  test("option") {
+    case class Foo(field: Option[String]) derives YamlCodec
+
+    val some = Foo(Some("some"))
+    val none = Foo(None)
+
+    assertNoDiff(some.asYaml, "field: some")
```
tgodzik, line 134: Do we have a test going the other way? Or printing is not possible at all currently?

tgodzik (approved):
LGTM, just one comment

core/shared/src/main/scala/org/virtuslab/yaml/YamlEncoder.scala
```diff
@@ -18,6 +18,11 @@ object YamlEncoder extends YamlEncoderCrossCompanionCompat {
   implicit def forBoolean: YamlEncoder[Boolean] = v => Node.ScalarNode(v.toString)
   implicit def forString: YamlEncoder[String]   = v => Node.ScalarNode(v)
 
+  implicit def forOption[T](implicit encoder: YamlEncoder[T]): YamlEncoder[Option[T]] = {
+    case Some(t) => encoder.asNode(t)
+    case None    => Node.ScalarNode("", Tag.nullTag)
```
kpodsiad, line 23: is there some reason to encode null as scalar node with null tag? I know this should be question to myself, but just thinking loudly

lbialy: There are no Null nodes and I didn't check if `null` can be passed as a value to Scalar node. I wouldn't expect it to be tbh.

kpodsiad: `Node.ScalarNode("", Tag.nullTag)` seems like a hack to represent something more complex that model was designed for :D It isn't obvious for me, but maybe it is ok because it is how spec defines it - https://yaml.org/spec/1.2.2/#10211-null.

kpodsiad: ehh, I now saw the how `Tag` looks like. I think it's ok as it is tag which determines how value of scalar should be interpreted

core/shared/src/test/scala-3/org/virtuslab/yaml/BaseYamlEncoderSuite.scala
lbialy: Oh, sure it is!

tgodzik:
@lbialy I think this can be merged?

lbialy:
Wait, tests!

lbialy:
I went through this again and there are separate Option decoder tests, we do miss circular tests but I don't think they should be part of this PR. This can be merged.

PR #299, opened 2024-05-29
Fix parsing numbers

sideeffffect:
Use java.lang.Integer.decode which handles other radices besides 10

Also remove `_` from the number string

### DIFF
--- a/core/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
@@ -74,15 +74,17 @@ object YamlDecoder extends YamlDecoderCompanionCrossCompat {
             case Tag.nullTag => Right(None)
             case Tag.boolean => value.toBooleanOption.toRight(cannotParse(value, "Boolean", node))
             case Tag.int =>
-              if (value.startsWith("0b"))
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
```diff
@@ -74,15 +74,17 @@ object YamlDecoder extends YamlDecoderCompanionCrossCompat {
             case Tag.nullTag => Right(None)
             case Tag.boolean => value.toBooleanOption.toRight(cannotParse(value, "Boolean", node))
             case Tag.int =>
-              if (value.startsWith("0b"))
-                Try(Integer.parseInt(value.drop(2), 8)).toEither.left
-                  .map(t => ConstructError.from(t, "Int", node))
-              else if (value.startsWith("0x"))
-                Try(Integer.parseInt(value.drop(2), 8)).toEither.left
```
sideeffffect, line 81: 8 instead of 16.

But it's better to leave it to `Long.decode` anyway.

tgodzik (commented):
Thanks for the contribution! I see that there is a failed test, could you take a look at it?

tgodzik (approved):
LGTM! Thanks!

PR #300, opened 2024-05-29
Make YamlError an Exception, makes for easier interop

sideeffffect:


lbialy:
A little bit left there, some tests to check positionality of errors. I'll
get to this.

sideeffffect:
So technically there may result conflicts with these two PRs.

In spirit they are not in conflict, right? They don't go against each other, if I understood it correctly. Nor do I have any problem with the aim of that other PR.

But this PR is much simpler and can be merged and released right now, which would unblock me in a downstream project. Would it be possible to merge and release this toady, please @tgodzik @lbialy  ?

lbialy:
Let me take a quick look.

lbialy (approved):

lbialy:
yeah, it's fine

lbialy:
@tgodzik will you handle the release or should I? do we do snapshots here?

tgodzik:
Versions 0.1.0 is running, but we do release snapshots.

PR #303, opened 2024-05-31
Add helper methods to Yaml{Encodr, Decoder, Codec}

sideeffffect:


sideeffffect:
One more :slightly_smiling_face: @lbialy @tgodzik

core/shared/src/main/scala/org/virtuslab/yaml/YamlCodec.scala
```diff
@@ -3,6 +3,25 @@ package org.virtuslab.yaml
...
  * A type class that provides both-way conversion between [[Node]] and [[T]]
  */
-trait YamlCodec[T] extends YamlDecoder[T] with YamlEncoder[T]
+trait YamlCodec[T] extends YamlDecoder[T] with YamlEncoder[T] { self =>
 
-object YamlCodec extends YamlCodecCompanionCrossCompat
+  def imap[T1](f: T => T1)(g: T1 => T): YamlCodec[T1] =
+    YamlCodec.from(self.map(f), self.contramap(g))
+
+  def iemap[T1](f: T => Either[ConstructError, T1])(g: T1 => T): YamlCodec[T1] =
```
lbialy, line 11: is there a better name for this shape of iso map? iemap is actively hostile (imap is too, tbh) if you're not familiar with ct and/or haskell

lbialy (changes requested):
iemap will need a better name, I think

lbialy:
Some tests would be nice too, I know that it's basically _if it compiles, it works_ kind of a change but still - let's say it's just for breakage detection

lbialy (approved):

PR #311, opened 2024-07-11
fixed issue 281 but is this correct?

### LINKED ISSUE #281: Parsing multiline string with escaped new line
If I have attribute 
```
description: |-
  Hi
  my name
  is John
```
I expect the value will be 
```
Hi
my name
is John
```
and not 
```
Hi\nmy name\nis John
``` 
The new lines should not be escaped `"\\n"`.

Tested with version 0.0.8

lbialy:
This change fixes #281 buuuut there's a bunch of tests that fail now, like:

```
=> Diff (- obtained, + expected)
     Scalar(
-      value = """# The ∙
-CRARG
-# We
-""",
+      value = "# The \\nCRARG\\n# We\\n",
       style = Literal(),
```

```
=> Diff (- obtained, + expected)
     Scalar(
-      value = """-----BEGIN CERTIFICATE-----
-0MTk0MVoXDenkKThvP7IS9q
-+Dzv5hG392KWh5f8xJNs4LbZyl901MeReiLrPH3w=
------END CERTIFICATE----""",
+      value = "-----BEGIN CERTIFICATE-----\\n0MTk0MVoXDenkKThvP7IS9q\\n+Dzv5hG392KWh5f8xJNs4LbZyl901MeReiLrPH3w=\\n-----END CERTIFICATE----",
       style = Literal(),
```

So apparently tests were written **with** the assumption that newlines will end up transformed to escaped newlines: `\n`. I don't think this is correct and I think the author of #281 is correct that we should not tamper with strings and provide them as they are for literals.

### DIFF
--- a/core/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/token/ScalarStyle.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/internal/load/reader/token/ScalarStyle.scala
@@ -11,6 +11,7 @@ object ScalarStyle {
   def escapeSpecialCharacter(scalar: String, scalarStyle: ScalarStyle): String =
     scalarStyle match {
       case ScalarStyle.DoubleQuoted => scalar
+      case ScalarStyle.Literal      => scalar
       case _ =>
         scalar.flatMap { char =>
           char match {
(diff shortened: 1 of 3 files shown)


lbialy:
@kpodsiad any idea why this was written this way?

lbialy:
I've tested that with circe-yaml:
```scala
//> using dep "io.circe::circe-yaml:1.15.0"

import io.circe.yaml

@main def circeYamlTest =
  val yml = """|description: |-
               |  Hi
               |  my name
               |  is John""".stripMargin

  yaml.parser.parse(yml) match
    case Left(err) => throw err
    case Right(value) =>
      println(value.hcursor.downField("description").as[String].getOrElse(???))
```

and it does keep newlines unescaped:
```
λ scala-cli run .                                                                                                                                                                                                             
Hi
my name
is John
```

so we're going to preserve them too.

tgodzik (approved):
Makes sense, LGTM

PR #315, opened 2024-07-16
parse all edge cases of ints and floats

### LINKED ISSUE #222: Hexadecimal values or `.inf` fails to get parsed
Yaml values such as `0xFFFF` or `-.inf` are recognized as `Integer` and `Double` but parsing fails:

```scala
scala> println("0xFFFF".as[Int])
Left(ConstructError(Cannot parse 0xFFFF as Int
at 0:0, expected Int
0xFFFF
^ ))

scala> println("-.inf".as[Double])
Left(ConstructError(Cannot parse -.inf as Double
at 0:0, expected Double
-.inf
^ ))
```

lbialy:
closes #222

### DIFF
--- a/core/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
@@ -100,43 +100,29 @@ object YamlDecoder extends YamlDecoderCompanionCrossCompat {
       .map(ConstructError.from(_, "Long", s))
   }
 
-  private val infinityRegex = """([-+]?)(\.inf|\.Inf|\.INF)""".r
-
-  private val nanRegex = """(\.nan|\.NaN|\.NAN)""".r
-
   implicit def forDouble: YamlDecoder[Double] = YamlDecoder { case s @ ScalarNode(value, _) =>
-    infinityRegex.findFirstMatchIn(value) match {
-      case Some(m) =>
-        Right(m.group(1) match {
-          case "-" => Double.NegativeInfinity
-          case _   => Double.PositiveInfinity
-        })
-      case None =>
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
```diff
@@ -85,27 +85,66 @@ object YamlDecoder extends YamlDecoderCompanionCrossCompat {
...
   implicit def forLong: YamlDecoder[Long] = YamlDecoder { case s @ ScalarNode(value, _) =>
-    Try(java.lang.Long.decode(value.replaceAll("_", "")).toLong).toEither.left
+    val normalizedValue =
+      if (value.startsWith("0o")) value.stripPrefix("0o").prepended('0') else value
+
+    Try(java.lang.Long.decode(normalizedValue.replaceAll("_", "")).toLong).toEither.left
       .map(ConstructError.from(_, "Long", s))
   }
 
+  private val infinityRegex = """([-+]?)(\.inf|\.Inf|\.INF)""".r
```
tgodzik, line 103: Do we need a regex? toLowerCase.endsWith etc. wouldn't be enough here?

lbialy: according to spec `.INf` is not a valid`Infinity` :[

also it would be something like:
```scala
if value.nonEmpty then
  if value(0) == '-' then
    if value(1) == '.' then
      if value(2) == 'i' || value(2) == 'I' ...
 ```
which is basically the same thing as what regex would do.

tgodzik: Ok sure, that was just a nitpick

tgodzik (approved):

PR #326, opened 2024-07-21
added handling for default parameters

### LINKED ISSUE #215: Use default arguments when field is not provided
`Upickle` and some other `json` libraries uses default arguments when field is not provided. It'll highly improve readability of `yaml` when you don't need to provide optional fields, and will give Devs much more flexibility.

```scala

//> using scala "3.2.2"
//> using dep "org.virtuslab::scala-yaml:0.0.6"

import org.virtuslab.yaml.*

case class Foo(id:String, name:String = "optional") derives YamlCodec

val example1 = 
	"id: 0112\n" + 
	"name: Tottally not fake name\n"
	
val example2 = 
	"id: 0113\n"

println(">" + example1.as[Foo]) //> Right(Foo(0112,Tottally not fake name))
println(">" + example2.as[Foo]) //> Left(ConstructError(Key name doesn't exist in parsed document)) 
// expected                     //> Right(Foo(0113,optional))
```

lbialy:
had to go from dumb mirrors to a full blown macro for products but hey, mirrors do not expose default params

closes #215

### DIFF
--- a/core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
+++ b/core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
@@ -43,7 +43,7 @@ object DecoderMacros {
   protected def constructValues[T](
       instances: List[(String, YamlDecoder[?], Boolean)],
       valuesMap: Map[String, Node],
-      defaultParams: Map[String, Any],
+      defaultParams: Map[String, () => Any],
       p: Mirror.ProductOf[T],
       parentNode: Node
   ): Either[ConstructError, T] = {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
```diff
@@ -94,17 +36,153 @@ private[yaml] trait DecoderMacros {
...
+          val typeArgs = TypeRepr.of[T].typeArgs
+          val identsExpr: Expr[List[() => Any]] =
+            if typeArgs.isEmpty then
+              Expr.ofList(idents.map(_.asExpr).map { case '{ $x } => '{ () => $x } })
+            else
+              Expr.ofList(
+                idents.map(_.appliedToTypes(typeArgs).asExpr).map { case '{ $x } => '{ () => $x } }
+              )
 
+          '{ $namesExpr.zip($identsExpr).toMap }
```
prolativ, line 182: I'm not sure if the order of declarations in `body` is reliable. Maybe it would be better to use something like `sym.companionClass.methodMember("$lessinit$greater$default$" + n.toString).head` to get the default value provider method for the n-th field?

prolativ (approved):

PR #327, opened 2024-07-21
improved integration tests a little bit

lbialy:


.github/workflows/ci.yml
```diff
@@ -61,8 +61,6 @@ jobs:
         env:
           PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
           PGP_SECRET: ${{ secrets.PGP_SECRET }}
-          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
```
tgodzik, line 64: This should stay, otherwise not password and username would be available. The release should be fixed, it seems org secrets were not set for this repo, which is why it didn't see the change

tgodzik (changes requested):

.github/workflows/ci.yml
lbialy: reverted

PR #328, opened 2024-07-24
rewritten to be a full macro to fix compiler issue

lbialy:
ffs

### DIFF
--- a/core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
+++ b/core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
@@ -6,39 +6,50 @@ import scala.compiletime.*
 import scala.quoted.*
 import scala.deriving.Mirror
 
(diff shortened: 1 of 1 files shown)


lbialy:
@prolativ can you please ack this?

docs/_docs/examples.md
```diff
@@ -28,7 +28,7 @@ key3:
 
 (there is an [issue](YamlDecoder) with deriving YamlEncoder instance for Option datatype hence `YamlDecoder` instead `YamlCodec`)  
 
-```scala sc:compile
```
tgodzik, line 31: So this is no longer checked in the docs?

tgodzik (approved):

docs/_docs/examples.md
lbialy: No, can't be temporarily. Scaladoc has an issue with that.

tgodzik: Ok, it seems it's just a scaladox issues. @lbialy please raise it upstream

PR #334, opened 2024-07-30
issue 331 - preserve order of keys when parsing and printing

### LINKED ISSUE #331: Preserve order?
In the following code the properties are printed in different order:
```
import org.virtuslab.yaml.*

val yaml = """
P:
  a: 0
  b: 1
  c: 2
  d: 3
  e: 4
"""

val node = yaml.asNode.toOption.get

node.asYaml
```

While property order is not mandated by YAML (or JSON), it is a common extension to support it (e.g. https://github.com/circe/circe-yaml/pull/23/files). Would it be possible to implement preserving property order in scala-yaml?

lbialy:
closes #331 

@sideeffffect @OndrejSpanel any comments?

### DIFF
--- a/core/shared/src/main/scala/org/virtuslab/yaml/Node.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/Node.scala
@@ -23,6 +23,13 @@ sealed trait Node {
 }
 
 object Node {
+
+  // Only for MappingNode key order, only ScalarNodes can be keys.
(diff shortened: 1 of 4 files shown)


OndrejSpanel (approved):
Looks great to me.

tgodzik (approved):

prolativ (approved):

PR #369, opened 2024-12-23
Support for parsing multiple documents

coreyoconnor:
This adds  `parseAllYamls`,  `parseYaml`, and the `asMany` extension method to the `yaml` package object.

- `parseAllYamls` either fails or returns a list of all yaml documents (`Node`) contained in the string content.
- `parseYaml` is an explicit form of the is the single document parsing provided by `.asNode`. 
- String extension method `asMany[T]` - decodes string of multiple documents into a list of `T`

To implement this, a fairly small addition was made to `Composer`: `multipleFromEvents`. Which uses `composeNode` to parse `Node`s from the events so long as there are `tail` events returned.

This also exposes the `pos` method of `Node`. This is useful when writing custom decoders for nicer error messages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   scalafmt:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v4
@@ -22,7 +22,7 @@ jobs:
       - run: sbt scalafmtCheckAll
 
   test:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
@@ -48,7 +48,7 @@ jobs:
   publish:
     needs: test
     if: github.event_name != 'pull_request'
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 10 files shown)


tgodzik:
@lbialy could you take a quick look?

coreyoconnor:
Thanks for the review! Made the style changes and added a minimal test.

lbialy:
Could you please rebase? I think the missing commits that touched sbt conf might be causing CI to fail (no idea why though :/ ). 

Also, wouldn't it also make sense to add `.asMany[T]` combinator in this PR?

coreyoconnor:
Sure! I'll do both

Looks like CI got the updated ubuntu image which is missing sbt. I'll
select the prior release as well

coreyoconnor:
@lbialy Done with the latest set of changes. Given the term `asMany` uses "many" for plural I changed the `parseAllYamls` to `parseManyYamls`

There is also style change I like, but feel free to request me to undo: Have the String extension methods forwards to top level methods: `decodeYaml`, `decodeManyYamls` etc.

coreyoconnor:
fixing...

coreyoconnor:
what a strange one. The implicit string conversion in base suite conflicted with the one in the package object. Yet the compile was fine with this - until I did a clean recompile.

tgodzik (approved):
I fixed the CI isues in main and reverted here. LGTM from me, unless @lbialy objects I will merge tomorrow.

lbialy (approved):

tgodzik:
Thanks @coreyoconnor !

Issue #367, opened 2024-12-10
Extra blank line when serializing mapping inside of sequence


bishabosha:
when the serializer does "push and indent" for entering mapping/sequence, there is always a newline inserted [see code](https://github.com/VirtusLab/scala-yaml/blob/002a0565d229a7416e733d60e7990de53a709daa/core/shared/src/main/scala/org/virtuslab/yaml/internal/dump/present/PresenterImpl.scala#L89), which leads to extra blank lines.

e.g. `Map("foo" -> Seq(Map("bar" -> 123))).asYaml`

### Output

```yaml
foo: 
  - 
    bar: 123
```

### Expectation

```yaml
foo: 
  - bar: 123
```

It would be nice to avoid the extra blank line specifically when the first entry of a sequence is a mapping. (it is fine for scalar values)

perhaps this could be configurable if there is a possibility of conflict

bishabosha:
this function works to post-process the string afterwards:
```scala
def shiftLines(yaml: String): String =
  val lines = yaml.linesIterator.toVector
  def startsWith[A](a: A, it: Iterator[A]): Boolean = it.hasNext && it.next == a
  lines.tail
    .foldLeft(lines.head :: Nil) { (before, line) =>
      val prev :: rest = before: @unchecked
      val prevEnd = prev.reverseIterator.dropWhile(_.isWhitespace)
      val lineStart = line.iterator.dropWhile(_.isWhitespace)
      if startsWith('-', prevEnd) && !startsWith('#', lineStart) then
        val (padding, content) = line.span(_.isWhitespace)
        (prev.padTo(padding.size, ' ') ++ content) :: rest
      else line :: prev :: rest
    }
    .reverse
    .mkString("\n")
```

coreyoconnor:
This adds an extra bit of challenge when doing development for me: My editor always strips trailing spaces on save. Which must be disabled to work with the scala-yaml codebase. Otherwise the editor will strip trailing spaces from test cases. Ouch!

PR #374, opened 2025-01-19
fix(#367): Do not add an extra space on output

coreyoconnor:
WIP only cause this is done on top of my other branch. Can be separated just lazy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   scalafmt:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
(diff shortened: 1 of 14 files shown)


bishabosha:
do you still intend to fix this - do you want some help?

coreyoconnor:
I have no plans to fix this at this time. Totally forgot about this one actually. Feel free to replace this PR with your own if you are game. Otherwise I'll see about cycling back to this after I do some other open source work. 😊

lbialy:
I think the PR @plokhotnyuk opened will be taking care of the same concern.

coreyoconnor:
@lbialy ah! I'll close this one then :)

PR #393, opened 2025-05-08
Fix Recursive Case Class Decoding

Dolfik1:
This commit fixes an issue with decoding recursive case classes in the library. The patch addresses a bug where recursive case classes failed to be parsed correctly from YAML. A test case for a Tree trait with Branch and Leaf case classes was added to demonstrate the fix.

Although I'm not an expert in Scala or metaprogramming, I worked through the problem and implemented a solution based on my understanding.

### DIFF
--- a/core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
+++ b/core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
@@ -17,34 +17,51 @@ object DecoderMacros {
       case '{ $m: Mirror.ProductOf[T] } => deriveProduct(m)
       case '{ $m: Mirror.SumOf[T] }     => deriveSum(m)
 
-  protected def summonSumOf[T <: Tuple: Type](using q: Quotes): List[Expr[YamlDecoder[_]]] =
+  protected def summonSumOf[T <: Tuple: Type](using q: Quotes): List[Expr[YamlDecoder[_]]] = {
     import q.reflect.report.*
     Type.of[T] match
       case '[t *: ts] =>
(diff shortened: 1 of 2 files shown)


tgodzik:
Thanks for the contribution! @lbialy should be able to review, but might be later this month.

core/shared/src/test/scala-3/org/virtuslab/yaml/decoder/DecoderSuite.scala
```diff
@@ -577,3 +577,39 @@ class DecoderSuite extends munit.FunSuite:
       case Right(value) =>
         fail(s"Should fail, but got $value")
   }
+
+  test("Recursive case class".only) {
```
tgodzik, line 581:
```suggestion
  test("Recursive case class") {
```
this fails on CI

core/shared/src/main/scala-3/org/virtuslab/yaml/YamlDecoderCrossCompat.scala
```diff
@@ -17,34 +17,51 @@ object DecoderMacros {
...
       case '{
-            type elementTypes <: Tuple;
+            type elementTypes <: Tuple
             $m: Mirror.SumOf[T] { type MirroredElemTypes = `elementTypes` }
           } =>
         val instancesExpr = Expr.ofList(summonSumOf[elementTypes])
         '{
           new YamlDecoder[T] {
-            private val instances = $instancesExpr.asInstanceOf[List[YamlDecoder[T]]]
+            private val instances = $instancesExpr.asInstanceOf[List[YamlDecoder[Any]]]
```
tgodzik, line 58: Do we need to change it here?

tgodzik (commented):
Some initial comments while we wait for some review from @lbialy

PR #399, opened 2025-06-02
Use Scaladex badge on ReadMe to show Scala version support

rtyley:
This Scaladex badge summarises which versions of Scala are supported by `scala-yaml` (and what the latest `scala-yaml` version is for each of those Scala versions):

[scala-yaml Scala version support](https://index.scala-lang.org/virtuslab/scala-yaml/scala-yaml)

More details on the badge format: https://github.com/scalacenter/scaladex/pull/660

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 ![Example](./docs/_assets/Scala_YAML_red.svg)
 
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Thanks!

PR #478, opened 2026-06-19
fix: decode non-finite floats as Any

### LINKED ISSUE #477: YamlDecoder[Any] rejects YAML non-finite float scalars
## Description

`YamlDecoder[Any]` rejects YAML non-finite float scalars such as `.inf`, `-.Inf`, `+.INF`, `.NAN`, `.nan`, and `.NaN`.

Direct `Float` and `Double` decoding already supports these spellings, but `Any` decoding goes through `forFloatPrecise.orElse(forDoublePrecise).orElse(forBigDecimal)`. The precise decoders parse the value as a non-finite float/double, compare `toString` (`Infinity`, `-Infinity`, or `NaN`) with the original YAML spelling, reject it, and then `BigDecimal` fails.

This looks like a remaining edge case after #330, which fixed the direct Float/Double NaN/Inf tag handling.

## Reproduction

```scala
import org.virtuslab.yaml.*

val yaml = "values: [.inf, -.Inf, +.INF, .NAN, .nan, .NaN]"
yaml.as[Any]
```

## Actual behavior

The call returns a `ConstructError` from the BigDecimal fallback, for example:

```text
Character i is neither a decimal digit number, decimal point, nor "e" notation exponential mark.
at 0:9, expected BigDecimal
values: [.inf, -.Inf, +.INF, .NAN, .nan, .NaN]
         ^
```

## Expected behavior

`YamlDecoder[Any]` should decode the scalars consistently with direct Float/Double decoding. Since `forAny` already prefers `Float` for float scalars that round-trip precisely, these values should decode as `Float.PositiveInfinity`, `Float.NegativeInfinity`, and `Float.NaN`.

## Validation

I have a small PR ready that:

- allows the precise Float/Double decoders to accept only the YAML non-finite spellings already recognized by `Tag`;
- adds a regression test for `yaml.as[Any]` on non-finite scalars;
- keeps invalid values like `-XXXinf` rejected by the direct Float decoder.

Local validation on latest `main`:

- `sbt core/test`
- `sbt coreJS/test`
- `sbt coreNative/test`
- `sbt '++2.13.18' core/test coreJS/test coreNative/test`
- `sbt scalafmtCheckAll`

He-Pin:
## Motivation

`YamlDecoder[Any]` rejects YAML non-finite float scalars such as `.inf`, `-.Inf`, `+.INF`, `.NAN`, `.nan`, and `.NaN`, even though direct `Float` and `Double` decoding already accepts these spellings.

The `Any` path uses the precise Float/Double decoders before falling back to `BigDecimal`. For non-finite scalars, the numeric value is parsed successfully, but `toString` returns `Infinity`, `-Infinity`, or `NaN`, which does not match the original YAML spelling. The precise decoder rejects it and `BigDecimal` then fails.

Related history: #330 fixed direct Float/Double NaN/Inf tag handling, but did not cover this `Any` precision path.

Closes #477.

## Modification

- Add a small helper that recognizes the YAML non-finite float spellings already accepted by `Tag`.
- Let `forFloatPrecise` and `forDoublePrecise` accept those non-finite spellings after successful numeric decoding.
- Add a regression test for `yaml.as[Any]` with non-finite float scalars.

## Result

`yaml.as[Any]` now decodes non-finite float scalars consistently with direct `Float`/`Double` decoding. Invalid values such as `-XXXinf` are still rejected because they do not match the existing `Tag` patterns.

Local validation:

- `sbt 'core/testOnly org.virtuslab.yaml.decoder.DecoderSuite'`
- `sbt core/test`
- `sbt coreJS/test`
- `sbt coreNative/test`
- `sbt '++2.13.18' core/test coreJS/test coreNative/test`
- `sbt scalafmtCheckAll`
- `git diff --check`

## Risk

Low. The change only affects values that already match the YAML non-finite float tag patterns and already decode successfully through the direct Float/Double decoders. Finite precise decoding and invalid non-finite-like strings keep the existing behavior.

### DIFF
--- a/core/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/YamlDecoder.scala
@@ -120,10 +120,14 @@ object YamlDecoder extends YamlDecoderCompanionCrossCompat {
     }
   }
 
+  private def isNonFiniteFloat(value: String): Boolean =
+    Tag.nan.matches(value) || Tag.plusInfinity.matches(value) || Tag.minusInfinity.matches(value)
+
   def forDoublePrecise: YamlDecoder[Double] = YamlDecoder { case s @ ScalarNode(value, _) =>
     forDouble.construct(s).flatMap { n =>
       val ns = n.toString
-      if (ns == value) Right(n) else Left(ConstructError.from(s"Double, decoded $ns", s))
+      if (ns == value || isNonFiniteFloat(value)) Right(n)
+      else Left(ConstructError.from(s"Double, decoded $ns", s))
     }
   }
 
@@ -143,7 +147,8 @@ object YamlDecoder extends YamlDecoderCompanionCrossCompat {
   def forFloatPrecise: YamlDecoder[Float] = YamlDecoder { case s @ ScalarNode(value, _) =>
     forFloat.construct(s).flatMap { n =>
       val ns = n.toString
-      if (ns == value) Right(n) else Left(ConstructError.from(s"Float, decoded $ns", s))
+      if (ns == value || isNonFiniteFloat(value)) Right(n)
+      else Left(ConstructError.from(s"Float, decoded $ns", s))
     }
   }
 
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

PR #488, opened 2026-07-16
More efficient parsing of YAML

plokhotnyuk:
@tgodzik It speeds up YAML parsing by about 6x times! 🚀 

Could you please review, merge, and cut a release ASAP? 

It will save your favorite performance-obsessed minion! 🏃‍♂️💨

### DIFF
--- a/core/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
@@ -40,22 +40,42 @@ object Tag {
   private[yaml] val plusInfinity  = "\\+?(\\.inf|\\.Inf|\\.INF)".r
   private[yaml] val nan           = "\\.nan|\\.NaN|\\.NAN".r
 
-  def resolveTag(value: String, style: Option[ScalarStyle] = None): Tag = {
-    val assumeString = style.exists(s => s == DoubleQuoted || s == SingleQuoted)
-    value match {
-      case null              => nullTag
-      case _ if assumeString => str
-      case nullPattern(_*)   => nullTag
-      case falsePattern(_*)  => boolean
-      case truePattern(_*)   => boolean
-      case int10Pattern(_*)  => int
-      case int8Pattern(_*)   => int
-      case int16Pattern(_*)  => int
-      case floatPattern(_*)  => float
-      case minusInfinity(_*) => float
-      case plusInfinity(_*)  => float
-      case nan(_*)           => float
-      case _                 => str
+  def resolveTag(value: String, style: Option[ScalarStyle] = None): Tag =
+    if (value eq null) nullTag
+    else if (style.exists(s => s == DoubleQuoted || s == SingleQuoted) || value.isEmpty) str
+    else {
+      value.charAt(0) match {
+        case 'n' | 'N' | '~' =>
+          if (nullPattern.matches(value)) nullTag
+          else str
+        case 'f' | 'F' =>
+          if (falsePattern.matches(value)) boolean
+          else str
+        case 't' | 'T' =>
+          if (truePattern.matches(value)) boolean
+          else str
+        case '-' =>
+          if (int10Pattern.matches(value)) int
+          else if (floatPattern.matches(value)) float
+          else if (minusInfinity.matches(value)) float
+          else str
+        case '+' =>
+          if (int10Pattern.matches(value)) int
+          else if (floatPattern.matches(value)) float
+          else if (plusInfinity.matches(value)) float
+          else str
+        case '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' =>
+          if (int10Pattern.matches(value)) int
+          else if (int8Pattern.matches(value)) int
+          else if (int16Pattern.matches(value)) int
+          else if (floatPattern.matches(value)) float
+          else str
+        case '.' =>
+          if (floatPattern.matches(value)) float
+          else if (plusInfinity.matches(value)) float
+          else if (nan.matches(value)) float
+          else str
+        case _ => str
+      }
     }
-  }
 }
--- a/core/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
@@ -42,7 +42,8 @@ object Tag {
 
   def resolveTag(value: String, style: Option[ScalarStyle] = None): Tag =
     if (value eq null) nullTag
-    else if (style.exists(s => s == DoubleQuoted || s == SingleQuoted) || value.isEmpty) str
+    else if (style.exists(s => s == DoubleQuoted || s == SingleQuoted)) str
+    else if (value.isEmpty) nullTag
     else {
       value.charAt(0) match {
         case 'n' | 'N' | '~' =>
--- a/core/shared/src/main/scala/org/virtuslab/yaml/YamlError.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/YamlError.scala
@@ -10,7 +10,7 @@ import org.virtuslab.yaml.internal.load.TagValue
 /**
  * An ADT representing a decoding failure.
  */
-sealed abstract class YamlError(val msg: String) extends Exception(msg)
+sealed abstract class YamlError(val msg: String) extends Exception(msg) with NoStackTrace
 
 sealed abstract class ParseError(override val msg: String) extends YamlError(msg)
 object ParseError {
@@ -74,9 +74,7 @@ object ConstructError {
   def from(t: Throwable): ConstructError = from(t.getMessage, None, None)
 }
 
-sealed abstract class ScannerError(override val msg: String)
-    extends YamlError(msg)
-    with NoStackTrace
+sealed abstract class ScannerError(override val msg: String) extends YamlError(msg)
 object ScannerError {
   def from(obtained: String, got: Token): ScannerError = Obtained(obtained, got)
 
--- a/core/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
+++ b/core/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
@@ -3,6 +3,7 @@ package org.virtuslab.yaml.internal.load.compose
 import scala.annotation.tailrec
 import scala.collection.mutable
 import scala.collection.immutable.ListMap
+import scala.util.control.NoStackTrace
 
 import org.virtuslab.yaml.ComposerError
 import org.virtuslab.yaml.Node
@@ -27,9 +28,16 @@ object ComposerImpl extends Composer {
   // A lightweight mutable wrapper avoiding `Result` allocation per event.
   private class Context(var events: List[Event])
 
+  // Internal exception used to fast-fail without paying the stack-trace generation cost
+  private case class ComposerException(err: YamlError) extends RuntimeException with NoStackTrace
+
   override def fromEvents(events: List[Event]): Either[YamlError, Node] = events match {
     case Nil => new Left(ComposerError("No events available"))
-    case _   => composeNode(new Context(events), mutable.Map.empty)
+    case _ =>
+      try new Right(composeNode(new Context(events), mutable.Map.empty))
+      catch {
+        case ComposerException(err) => new Left(err)
+      }
   }
 
   override def multipleFromEvents(events: List[Event]): Either[YamlError, List[Node]] = {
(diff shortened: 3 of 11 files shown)


Ostrzyciel (approved):
Andriy asked me for a review – I went over the code and it does make sense to me, as far as I understand it :)

Please note that I'm unfamiliar with this codebase, so this is a very weak review.

core/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
```diff
@@ -40,22 +40,42 @@ object Tag {
...
-      case int8Pattern(_*)   => int
-      case int16Pattern(_*)  => int
-      case floatPattern(_*)  => float
-      case minusInfinity(_*) => float
-      case plusInfinity(_*)  => float
-      case nan(_*)           => float
-      case _                 => str
+  def resolveTag(value: String, style: Option[ScalarStyle] = None): Tag =
+    if (value eq null) nullTag
+    else if (style.exists(s => s == DoubleQuoted || s == SingleQuoted) || value.isEmpty) str
```
Gedochao: It looks to me like it introduces a regression for empty values.
```suggestion
    else if (style.exists(s => s == DoubleQuoted || s == SingleQuoted)) str
    else if (value.isEmpty) nullTag
```
I'd add a test for it, smth like:
```scala
test("resolveTag resolves empty plain value to nullTag") {
  assertEquals(Tag.resolveTag(""), Tag.nullTag)
}

test("resolveTag resolves empty quoted value to str") {
  assertEquals(Tag.resolveTag("", Some(ScalarStyle.DoubleQuoted)), Tag.str)
}
```

core/shared/src/main/scala/org/virtuslab/yaml/Tag.scala
```diff
@@ -40,22 +40,42 @@ object Tag {
...
-      case truePattern(_*)   => boolean
-      case int10Pattern(_*)  => int
-      case int8Pattern(_*)   => int
-      case int16Pattern(_*)  => int
-      case floatPattern(_*)  => float
-      case minusInfinity(_*) => float
-      case plusInfinity(_*)  => float
-      case nan(_*)           => float
-      case _                 => str
+  def resolveTag(value: String, style: Option[ScalarStyle] = None): Tag =
```
Gedochao, line 43: Some more tests for edge cases I'd add to gain confidence that the behaviour remains unchanged here (in this case, as far as I've checked, there seems to be no regression, but let's run this on CI):
```scala
test("resolveTag resolves octal integers to int") {
    assertEquals(Tag.resolveTag("0o7"), Tag.int)
    assertEquals(Tag.resolveTag("0o777"), Tag.int)
  }

  test("resolveTag resolves hex integers to int") {
    assertEquals(Tag.resolveTag("0xFF"), Tag.int)
    assertEquals(Tag.resolveTag("0x3A"), Tag.int)
    assertEquals(Tag.resolveTag("0xABCD"), Tag.int)
  }

  test("resolveTag resolves unsigned infinity to float") {
    assertEquals(Tag.resolveTag(".inf"), Tag.float)
    assertEquals(Tag.resolveTag(".Inf"), Tag.float)
    assertEquals(Tag.resolveTag(".INF"), Tag.float)
  }

  test("resolveTag resolves positive infinity to float") {
    assertEquals(Tag.resolveTag("+.inf"), Tag.float)
    assertEquals(Tag.resolveTag("+.Inf"), Tag.float)
    assertEquals(Tag.resolveTag("+.INF"), Tag.float)
  }

  test("resolveTag resolves negative infinity to float") {
    assertEquals(Tag.resolveTag("-.inf"), Tag.float)
    assertEquals(Tag.resolveTag("-.Inf"), Tag.float)
    assertEquals(Tag.resolveTag("-.INF"), Tag.float)
  }

  test("resolveTag resolves NaN to float") {
    assertEquals(Tag.resolveTag(".nan"), Tag.float)
    assertEquals(Tag.resolveTag(".NaN"), Tag.float)
    assertEquals(Tag.resolveTag(".NAN"), Tag.float)
  }

  test("resolveTag resolves mixed-case null to nullTag") {
    assertEquals(Tag.resolveTag("null"), Tag.nullTag)
    assertEquals(Tag.resolveTag("Null"), Tag.nullTag)
    assertEquals(Tag.resolveTag("NULL"), Tag.nullTag)
    assertEquals(Tag.resolveTag("~"), Tag.nullTag)
  }

  test("resolveTag resolves mixed-case true to boolean") {
    assertEquals(Tag.resolveTag("true"), Tag.boolean)
    assertEquals(Tag.resolveTag("True"), Tag.boolean)
    assertEquals(Tag.resolveTag("TRUE"), Tag.boolean)
  }

  test("resolveTag resolves mixed-case false to boolean") {
    assertEquals(Tag.resolveTag("false"), Tag.boolean)
    assertEquals(Tag.resolveTag("False"), Tag.boolean)
    assertEquals(Tag.resolveTag("FALSE"), Tag.boolean)
  }
```

Gedochao (commented):
I'd wait for @tgodzik's opinion before giving an okay on this, as I'm not too familiar with this codebase.
That said, I left some comments.
Since this is a performance improvement PR, it could also be valuable to include some benchmark results.

plokhotnyuk:
@Gedochao Thanks for the review! It was an amazing catch!

I fixed it in a subsequent commit and added test coverage for `Tag.resolveTag` function

plokhotnyuk:
@Gedochao Could you please run CI workflow and review the latest commits with performance fixes?

plokhotnyuk:
@Gedochao Could you please give the CI workflow one more run?

@tgodzik Would you be a gem and review, merge, and cut a release for this turbo-charged scala-yaml?

core/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
```diff
@@ -27,134 +25,148 @@ trait Composer {
...
 
   override def fromEvents(events: List[Event]): Either[YamlError, Node] = events match {
-    case Nil => Left(ComposerError("No events available"))
-    case _   => composeNode(events, mutable.Map.empty).map(_.node)
+    case Nil => new Left(ComposerError("No events available"))
+    case _ =>
+      try new Right(composeNode(new Context(events), mutable.Map.empty))
+      catch {
+        case ComposerException(err) => new Left(err)
+      }
```
Gedochao, line 40: why do we need to switch to a `try`-`catch` clause here, and throw `ComposerError`s in `composeNode`?  The old approach was cleaner and easier to read, returning `Left` for errors.

core/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
```diff
@@ -391,119 +352,108 @@ final class ParserImpl private (in: Tokenizer) extends Parser {
...
+    else if (p eq ParseFlowMappingEntryOpt) parseFlowMappingEntryOpt(token)
+    else if (p eq ParseFlowMappingComma) parseFlowMappingComma(token)
+    else if (p eq ParseFlowSeqEnd) parseFlowSeqEnd(token)
+    else if (p eq ParseFlowSeqEntry) parseFlowSeqEntry(token)
+    else if (p eq ParseFlowSeqEntryOpt) parseFlowSeqEntryOpt(token)
+    else if (p eq ParseFlowSeqComma) parseFlowSeqComma(token)
+    else if (p eq ParseFlowSeqPairKey) parseFlowPairKey(token)
+    else if (p eq ParseFlowSeqPairValue) parseFlowPairValue(token)
+    else p.asInstanceOf[ReturnEvent].produceEvent(token)
+  }
```
Gedochao, line 458: I wonder if it's worth it to rewrite this to `if`-`else` chain here. IMO it isn't - what are we gaining here? We lose match exhaustiveness detection on compiler level - I'd expect `Production` is a sealed class to catch missing cases, here. Which we won't be getting with `if-else`

core/shared/src/main/scala/org/virtuslab/yaml/YamlError.scala
```diff
@@ -10,7 +10,7 @@ import org.virtuslab.yaml.internal.load.TagValue
 /**
  * An ADT representing a decoding failure.
  */
-sealed abstract class YamlError(val msg: String) extends Exception(msg)
+sealed abstract class YamlError(val msg: String) extends Exception(msg) with NoStackTrace
```
Gedochao, line 13: Won't it be a big UX regression, with errors effectively becoming way harder to debug?

core/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
```diff
@@ -27,134 +25,148 @@ trait Composer {
 }
 
 object ComposerImpl extends Composer {
-  private case class Result[+T](node: T, remaining: List[Event])
-  private type ComposeResult[+T] = Either[YamlError, Result[T]]
-  // WithPos is used in inner tailrec methods because they also return position of first child
-  private type ComposeResultWithPos[T] = Either[YamlError, (Result[T], Option[Range])]
+  // A lightweight mutable wrapper avoiding `Result` allocation per event.
+  private class Context(var events: List[Event])
```
Gedochao, line 29: I don't like the newly introduced mutability here, but I understand that's a big part of what we are gaining by saving on allocation...

Gedochao (commented):
Already looks quite good, although I still have some concerns, left a few comments.

core/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
plokhotnyuk: `if-else` just compare reference without overhead of `.equals` calls. If the new case object/class will be added without modification of this method it will throw an exception in `p.asInstanceOf[ReturnEvent]`

core/shared/src/main/scala/org/virtuslab/yaml/YamlError.scala
plokhotnyuk: No UX regression is expected because we do not throw `YamlError` for users only pass them as `Left[YamlError]` values. Moreover I would expected more efficiency and safety due to not collecting of stack traces, and clear understanding (less cognitive load) by avoiding mixing of stack-traces and non-stack-traces exception in the same sealed hierarchy of errors.

core/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
plokhotnyuk: Yes it saves a lot on allocations and the public method is reference transparent (without side effects) yet

core/shared/src/main/scala/org/virtuslab/yaml/internal/load/compose/Composer.scala
plokhotnyuk: `try-catch` on a `NoStackTrace` error saves a lot on redundant allocations for nested recursive calls during composing of AST nodes, while keeping referentially transparency for the public methods.

lbialy:
Looks good all in all, I was concerned about this perf work being done before parser is improved to handle round-trips (the current impl fails round-trips for huge amount of YAML cases) but all changes seem to be pure wasteful allocation removals and don't really modify the logic. +1 from me.

core/shared/src/main/scala/org/virtuslab/yaml/YamlError.scala
lbialy: not even a binary compat issue actually

lbialy: although it would be wise to review if the messages contain good positional feedback about what failed to parse where (in the future)

Gedochao (approved):
LGTM
I don't really like how it looks in the code, however. 
After we merge this, the code'll be a lot less readable and there's no check on the CI that would keep a maintainer from merging a performance regression which fixes the DX.
Setting up benchmarks is a separate story and out of scope here, of course.

core/shared/src/main/scala/org/virtuslab/yaml/internal/load/parse/ParserImpl.scala
```diff
@@ -113,276 +104,246 @@ final class ParserImpl private (in: Tokenizer) extends Parser {
...
-    case TokenKind.FlowSequenceStart =>
-      productions.prependAll(ParseFlowNode :: Nil)
-      getNextEventImpl()
-    case TokenKind.Scalar(_, _) | _: TokenKind.Anchor =>
-      productions.prependAll(
-        ParseFlowNode :: ParseFlowMappingComma :: ParseFlowMappingEntry :: Nil
-      )
+    case _: TokenKind.Scalar | _: TokenKind.Anchor =>
+      productions.prependAll(Array(ParseFlowNode, ParseFlowMappingComma, ParseFlowMappingEntry))
       parseFlowNode(token)
```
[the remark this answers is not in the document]
plokhotnyuk: I've addressed this potential issue in a subsequent commit

plokhotnyuk:
I've addressed issues found by Copilot.

In NeverBlink we have a comprehensive set of benchmarks for LinkML that is based on YAML:
https://github.com/NeverBlink-labs/linkml-benchmark-schemas

And a smaller one which I run on regular basis:
https://github.com/NeverBlink-OSS/linkml-scala/tree/main/benchmark

Currently the remaining performance issue is using of `ListMap` in the AST composer for `Node.MappingNode`, that lead to O(n^2) algorithmic complexity when accessing map node values by keys on the user side. Below is async-profiler's flame graph for CPU cycles burned which highlights that ~30% overhead by magenta color:

image

That is why I'm proposing to make `ParserImpl.getEvents` public in a subsequent commit.

lbialy:
@plokhotnyuk are you still working on this? I assume you're still working on that `ListMap` problem

plokhotnyuk:
@lbialy Unfortunately no, the issue is much broader... 

When I started to replace `ListMap` by `VectorMap` another problem pops up with inefficient `hashCode` calculation for nodes that makes code complexity even worse...

Let's stop at this level of optimization, have it merged and released and then I will try to solve it for existing node AST API or will propose new one.

lbialy (approved):
lgtm

plokhotnyuk:
@lbialy @Gedochao @tgodzik I was wrong about 6x speed up estimation...

On [real-world yaml files](https://github.com/NeverBlink-OSS/linkml-scala/pull/116) it shows 7x-10x speed up and ~20x times less allocation rate per operation!!!

org.virtuslab::scala-yaml::0.3.2
```
Benchmark                                                       (schema)   Mode  Cnt           Score          Error   Units
YamlParsingBench.fromStringToNode                         cgmes-core.yml  thrpt   15          72.361 ±        1.860   ops/s
YamlParsingBench.fromStringToNode:gc.alloc.rate           cgmes-core.yml  thrpt   15        8908.800 ±      227.009  MB/sec
YamlParsingBench.fromStringToNode:gc.alloc.rate.norm      cgmes-core.yml  thrpt   15   129113063.245 ±    28509.685    B/op
YamlParsingBench.fromStringToNode:gc.count                cgmes-core.yml  thrpt   15          58.000                 counts
YamlParsingBench.fromStringToNode:gc.time                 cgmes-core.yml  thrpt   15          34.000                     ms
YamlParsingBench.fromStringToNode                     cgmes-dynamics.yml  thrpt   15          14.930 ±        0.868   ops/s
YamlParsingBench.fromStringToNode:gc.alloc.rate       cgmes-dynamics.yml  thrpt   15        8737.397 ±      107.043  MB/sec
YamlParsingBench.fromStringToNode:gc.alloc.rate.norm  cgmes-dynamics.yml  thrpt   15   615661923.372 ± 40989267.773    B/op
YamlParsingBench.fromStringToNode:gc.count            cgmes-dynamics.yml  thrpt   15          60.000                 counts
YamlParsingBench.fromStringToNode:gc.time             cgmes-dynamics.yml  thrpt   15          99.000                     ms
YamlParsingBench.fromStringToNode                            TC57CIM.yml  thrpt   15           4.536 ±        0.105   ops/s
YamlParsingBench.fromStringToNode:gc.alloc.rate              TC57CIM.yml  thrpt   15        8684.199 ±      200.665  MB/sec
YamlParsingBench.fromStringToNode:gc.alloc.rate.norm         TC57CIM.yml  thrpt   15  2007960186.880 ±      198.114    B/op
YamlParsingBench.fromStringToNode:gc.count                   TC57CIM.yml  thrpt   15          63.000                 counts
YamlParsingBench.fromStringToNode:gc.time                    TC57CIM.yml  thrpt   15         189.000                     ms
```

org.virtuslab::scala-yaml::0.3.3 (after https://github.com/VirtusLab/scala-yaml/pull/488)
```
Benchmark                                                       (schema)   Mode  Cnt         Score       Error   Units
YamlParsingBench.fromStringToNode                         cgmes-core.yml  thrpt   15       676.537 ±     3.467   ops/s
YamlParsingBench.fromStringToNode:gc.alloc.rate           cgmes-core.yml  thrpt   15      3697.193 ±    34.221  MB/sec
YamlParsingBench.fromStringToNode:gc.alloc.rate.norm      cgmes-core.yml  thrpt   15   5731042.414 ± 35802.682    B/op
YamlParsingBench.fromStringToNode:gc.count                cgmes-core.yml  thrpt   15        25.000              counts
YamlParsingBench.fromStringToNode:gc.time                 cgmes-core.yml  thrpt   15        38.000                  ms
YamlParsingBench.fromStringToNode                     cgmes-dynamics.yml  thrpt   15       153.342 ±     0.802   ops/s
YamlParsingBench.fromStringToNode:gc.alloc.rate       cgmes-dynamics.yml  thrpt   15      3779.196 ±    19.792  MB/sec
YamlParsingBench.fromStringToNode:gc.alloc.rate.norm  cgmes-dynamics.yml  thrpt   15  25846109.228 ±    27.751    B/op
YamlParsingBench.fromStringToNode:gc.count            cgmes-dynamics.yml  thrpt   15        27.000              counts
YamlParsingBench.fromStringToNode:gc.time             cgmes-dynamics.yml  thrpt   15       133.000                  ms
YamlParsingBench.fromStringToNode                            TC57CIM.yml  thrpt   15        34.472 ±     0.420   ops/s
YamlParsingBench.fromStringToNode:gc.alloc.rate              TC57CIM.yml  thrpt   15      3024.374 ±    36.910  MB/sec
YamlParsingBench.fromStringToNode:gc.alloc.rate.norm         TC57CIM.yml  thrpt   15  92008251.061 ±    12.367    B/op
YamlParsingBench.fromStringToNode:gc.count                   TC57CIM.yml  thrpt   15        21.000              counts
YamlParsingBench.fromStringToNode:gc.time                    TC57CIM.yml  thrpt   15       366.000                  ms
```

PR #493, opened 2026-07-22
Fix serialization of string values for different styles and properly escape values that require double quoting + add benchmarks for parsing

plokhotnyuk:
It also fixed redundant type conversion for string values. 

I think it should depend on decoding guided by user's data structures (case class field type, not yaml parser)

I would prefer round-trip parsing and serialization without unexpected type conversions (currently "1.20" will be transformed to 1.2, "0.0000001" to 1e-7)...

BTW, the preserved feature of enforcing of null values to be !!null is questionable...


plokhotnyuk:
@tgodzik @lbialy Could you please review it?

lbialy:
I'm looking at this but it's all rather murky. Is this very urgent @plokhotnyuk? I'd like to, at least, run the spec tests to see if this PR is moving us in the right direction. I assume it is but with YAML spec being what it is, you actually never know.

plokhotnyuk:
Now it is not very urgent after I started to cut [perf-fork](https://github.com/plokhotnyuk/scala-yaml/tree/perf-fork) releases.

lbialy:
The spec suite didn't change between `main` and #493:
```
Passed: 118/322 36.65%
Failed: 204/322 63.35%
```

core/js/src/test/org/virtuslab/yaml/YamlEncoderSuite.scala
```diff
@@ -1,8 +1,11 @@
 package org.virtuslab.yaml
 
 import org.virtuslab.yaml.*
+import org.virtuslab.yaml.Node.*
+import org.virtuslab.yaml.internal.load.reader.token.ScalarStyle
 
 class YamlEncoderSpec extends munit.FunSuite:
```
lbialy: This whole class should be in core/shared/src/test, not in core/js but that's a different problem

lbialy (changes requested):
This change has to be YAML spec-compliant so I ran a verification study with AI:

# PR #493 Code Review: Scalar Serialization

## Overview

PR #493 introduces quoting logic in `PresenterImpl.serializeScalar` to ensure string-tagged scalars that would re-resolve as non-string types get double-quoted. The core invariant — quote `Tag.str` values whose lexical form matches boolean, null, or numeric patterns — is correct and fixes real bugs (e.g. `version: "3.9"` in docker-compose files). However, the implementation has several issues that violate the YAML 1.2.2 specification.

All issues below have been verified with failing tests in `core/shared/src/test/scala/org/virtuslab/yaml/presenter/PresenterBugsSuite.scala`.

## Spec Compliance Summary

| Bug | Failing Tests | YAML 1.2 Spec Violation? | Key Spec Section |
|---|---|---|---|
| **1: Trailing whitespace not quoted** | 4 | **Yes** | §7.3.3 — "All leading and trailing white space characters are excluded from the content" for plain scalars. Emitting `"true "` plain silently drops the space. |
| **2: Trailing colon not quoted** | 2 | **Yes** | §7.3.3 rule [130] — `:` in a plain scalar requires a following `ns-plain-safe` character. A trailing `:` (followed by newline/EOF) fails that lookahead and gets parsed as a mapping value indicator. |
| **4: Reference equality on tags** | 3 | **Yes** | §3.2.1.3 — tags must be compared by "simple character-by-character comparison." Using `eq` (object identity) causes non-singleton `CoreSchemaTag("tag:yaml.org,2002:str")` to bypass the quoting logic entirely. |
| **6: Folded style is lossy** | 1 | **Yes** | §6.5, §8.1.3 — in folded style single newlines between non-empty lines are converted to spaces (rule [72] `b-as-space`). The current `escapeFolded` is identical to `escapeLiteral`, so `"abc\ndef"` round-trips as `"abc def"`. |
| **7: Block scalars at indent 0** | 2 | **Yes** | §8.1.1.1, §6.1 — block scalar content indentation level is `n + m`, and the explicit indicator range is 1–9 (rule [163]). At `n=0`, content must still be indented at least 1 space. |
| **8: Non-scalar mapping key throws** | 1 | **Yes** | §3.2.1.1 — "keys may be arbitrary nodes." Complex keys are legal YAML and the `?` indicator exists for emitting them. Throwing a bare `RuntimeException` is non-compliant. |
| **Roundtrip: string "null"** | 1 | **Yes** | §3.2.1, §3.3.2, §10.3.2 — a `!!str` node whose content matches a non-string resolution pattern must be quoted or explicitly tagged to preserve type identity through round-trips. |

## Detailed Findings

### 1. Trailing whitespace check is gated by first-character dispatch

**File:** `PresenterImpl.scala`, `requiresDoubleQuoting`, lines 112–177

The leading/trailing whitespace check (`Character.isWhitespace`) lives inside the `case _` default branch of the first-character switch. When the first character matches an explicit case (`'t'`, `'n'`, `'f'`, `'0'`–`'9'`, etc.), the trailing whitespace check is never reached.

Affected values include any string with trailing whitespace whose first character hits a named branch: `"true "`, `"null "`, `"123 "`, `"3.14 "`, `"-foo "`, etc.

**Fix:** Move the leading/trailing whitespace and control character checks before the switch, as unconditional early returns.

### 2. Trailing colon not detected

**File:** `PresenterImpl.scala`, `requiresDoubleQuoting`, lines 112–177

The `':'` case at line 154 only checks `len == 1 || Character.isWhitespace(s.charAt(1))`, catching bare `:` and `: foo`. It does not catch strings that *end* with `:` (e.g. `"a:"`, `"host:"`). The inner loop at lines 166–175 checks `prev == ':' && c == ' '` but that only catches `: ` in the middle of a string, not a trailing colon.

**Fix:** Add a check for `s.charAt(len - 1) == ':'` as an early return condition.

### 3. `tag.exists(_ eq Tag.str)` uses reference equality

**File:** `PresenterImpl.scala`, `serializeScalar`, lines 101 and 108

Both tag comparisons use `eq` (JVM reference identity):
```scala
if (tag.exists(_ eq Tag.str)) { ... }
else if (tag.exists(_ eq Tag.nullTag)) sb.append("!!null")
```

`CoreSchemaTag` is a `case class`, so two instances with the same `value` are `==`-equal but not `eq`-equal. Any `Tag` constructed via a code path other than `Tag.str` / `Tag.nullTag` singletons (e.g. parsed from explicit `!!str`, constructed in user code) will fall through to the unprotected `else sb.append(value)` branch — no quoting, no escaping.

Additionally, `tag == None` (no tag at all) hits the same `else` branch, emitting values like `"true"` as raw plain scalars.

**Fix:** Replace `eq` with `==` or match on the tag's `value` string. Handle `tag == None` by running the quoting predicate defensively.

### 4. `escapeFolded` is identical to `escapeLiteral`

**File:** `PresenterImpl.scala`, `escapeFolded`, lines 252–282

The folded style implementation is a copy of `escapeLiteral` with `>` instead of `|`. But folded and literal have fundamentally different semantics: literal preserves all newlines, while folded converts single newlines to spaces. Emitting `"abc\ndef"` as:
```
>-
abc
def
```
causes a conformant parser to read it back as `"abc def"` — silent data corruption.

**Fix:** Either double newlines in folded output so they survive folding, emit more-indented lines (which are not folded), or fall back to a different style when the content contains single newlines that must be preserved.

### 5. Block scalars at top level have zero indentation

**File:** `PresenterImpl.scala`, `escapeLiteral`/`escapeFolded`, lines 220–282

When a block scalar is the top-level node, `indent` is 0. The content lines are emitted with no leading spaces, which means they are at the same indentation level as the block header — syntactically, they are not part of the block scalar. A conformant parser either errors or misinterprets the content.

Example output for `"abc\ndef"` at top level:
```
|-
abc
def
```

This is invalid — `abc` at column 0 is not indented relative to the block indicator.

**Fix:** Ensure block scalar content is always indented at least 1 space, even at the top level. When `indent == 0`, use an explicit indentation indicator (e.g. `|1` or `>1`) and indent content by 1 space.

### 6. `sys.error` on non-scalar mapping keys

**File:** `PresenterImpl.scala`, `serializeMapping`, line 58

```scala
case _ => sys.error("Cannot serialize non-scalar mapping key: " + head)
```

Complex mapping keys (sequences or mappings as keys) are legal YAML. The specification provides the `?` indicator for emitting them in block style. This `sys.error` throws a bare `RuntimeException` from library code, which is both a spec violation and a regression from the previous behavior.

**Fix:** Handle `MappingStart` and `SequenceStart` in the key position by emitting a `? ` prefix followed by the serialized complex key.

### 7. String "null" does not round-trip through encode/decode

This is an encoder-side issue rather than a presenter bug. `YamlEncoder.forString` creates `ScalarNode(escapeSpecialCharacters(v))`, which calls `Tag.resolveTag(value)`. For the string `"null"`, `resolveTag` returns `Tag.nullTag`, so the node loses its string identity. When serialized and re-parsed, it becomes a null value that cannot be decoded back as `String`.

The same class of issue affects `"true"`, `"false"`, `"3.9"`, `"~"`, etc. through the encoder path — though some of these happen to survive because the `String` decoder is tolerant of float/boolean-tagged scalars.

**Fix:** `YamlEncoder.forString` should always produce `ScalarNode` with `Tag.str`, not rely on `Tag.resolveTag` which is designed for the *parsing* path.

Below the contents of the test suite at `core/shared/src/test/scala/org/virtuslab/yaml/presenter/PresenterBugsSuite.scala`:

```
package org.virtuslab.yaml.presenter

import org.virtuslab.yaml._
import org.virtuslab.yaml.internal.dump.present.PresenterImpl
import org.virtuslab.yaml.internal.load.parse.EventKind._
import org.virtuslab.yaml.internal.load.parse.NodeEventMetadata
import org.virtuslab.yaml.internal.load.reader.token.ScalarStyle

class PresenterBugsSuite extends munit.FunSuite {

  private val nl = System.lineSeparator()

  private def presentScalar(
      value: String,
      tag: Tag = Tag.str,
      style: ScalarStyle = ScalarStyle.Plain
  ): String =
    PresenterImpl.asString(
      Seq(
        DocumentStart(),
        Scalar(value, style, NodeEventMetadata(tag = Some(tag))),
        DocumentEnd()
      )
    )

  // -----------------------------------------------------------------------
  // Bug 1: Trailing whitespace check is inside the `case _` default branch.
  //
  // The trailing-whitespace check in requiresDoubleQuoting only runs when
  // the first character doesn't match any explicit case. Strings like
  // "true " (first char 't') skip it, so trailing whitespace is not detected.
  // YAML strips trailing whitespace on plain scalars (§7.3.3), causing
  // data corruption.
  // -----------------------------------------------------------------------

  test("bug1: trailing whitespace on boolean-like string must be quoted") {
    val result = presentScalar("true ")
    assert(
      result.contains("\"true \""),
      s"Expected 'true ' to be double-quoted, but got: $result"
    )
  }

  test("bug1: trailing whitespace on number-like string must be quoted") {
    val result = presentScalar("123 ")
    assert(
      result.contains("\"123 \""),
      s"Expected '123 ' to be double-quoted, but got: $result"
    )
  }

  test("bug1: trailing whitespace on null-like string must be quoted") {
    val result = presentScalar("null ")
    assert(
      result.contains("\"null \""),
      s"Expected 'null ' to be double-quoted, but got: $result"
    )
  }

  test("bug1: trailing whitespace on float-like string must be quoted") {
    val result = presentScalar("3.14 ")
    assert(
      result.contains("\"3.14 \""),
      s"Expected '3.14 ' to be double-quoted, but got: $result"
    )
  }

  // -----------------------------------------------------------------------
  // Bug 2: Trailing colon not handled.
  //
  // "a:" is not quoted. Per §7.3.3 rule [130], `:` in a plain scalar
  // requires a following ns-plain-safe character. A trailing colon
  // (followed by newline/EOF) fails that lookahead and gets parsed as
  // a mapping value indicator.
  // -----------------------------------------------------------------------

  test("bug2: trailing colon must be quoted") {
    val result = presentScalar("a:")
    assert(
      result.contains("\"a:\""),
      s"Expected 'a:' to be double-quoted, but got: $result"
    )
  }

  test("bug2: trailing colon followed by nothing should be quoted") {
    val result = presentScalar("host:")
    assert(
      result.contains("\"host:\""),
      s"Expected 'host:' to be double-quoted, but got: $result"
    )
  }

  // -----------------------------------------------------------------------
  // Bug 4: `tag.exists(_ eq Tag.str)` uses reference equality.
  //
  // Per §3.2.1.3, tags must be compared by character-by-character string
  // comparison. A Tag constructed for "tag:yaml.org,2002:str" that isn't
  // the Tag.str singleton falls to the `else sb.append(value)` branch:
  // unquoted and unescaped.
  // -----------------------------------------------------------------------

  test("bug4: non-singleton str tag must still trigger quoting logic") {
    val nonSingletonStr = CoreSchemaTag("tag:yaml.org,2002:str")
    assert(
      nonSingletonStr != Tag.str || !(nonSingletonStr eq Tag.str),
      "Precondition: nonSingletonStr should not be reference-equal to Tag.str"
    )
    val result = presentScalar("true", tag = nonSingletonStr)
    assert(
      result.contains("\"true\""),
      s"Expected 'true' with non-singleton str tag to be double-quoted, but got: $result"
    )
  }

  test("bug4: non-singleton null tag must still trigger null rendering") {
    val nonSingletonNull = CoreSchemaTag("tag:yaml.org,2002:null")
    val result = presentScalar("", tag = nonSingletonNull)
    assert(
      result.contains("!!null"),
      s"Expected empty string with non-singleton null tag to render as !!null, but got: $result"
    )
  }

  test("bug4: scalar with no tag should still apply quoting for ambiguous values") {
    val events = Seq(
      DocumentStart(),
      Scalar("true", ScalarStyle.Plain, NodeEventMetadata.empty),
      DocumentEnd()
    )
    val result = PresenterImpl.asString(events)
    assert(
      !result.trim.equals("true"),
      s"Expected 'true' with no tag to be quoted or handled safely, but got: $result"
    )
  }

  // -----------------------------------------------------------------------
  // Bug 6: escapeFolded doesn't handle folded-style semantics.
  //
  // Per §6.5, §8.1.3, in folded style single newlines between non-empty
  // lines are converted to spaces (rule [72] b-as-space). The current
  // implementation is identical to literal style, so content with single
  // newlines will not round-trip correctly.
  // -----------------------------------------------------------------------

  test("bug6: folded style must round-trip single newlines correctly") {
    val input = "abc\ndef"
    val folded = presentScalar(input, style = ScalarStyle.Folded)
    val parsed = folded.as[String]
    parsed match {
      case Right(result) =>
        assertEquals(
          result,
          input,
          s"Folded scalar did not round-trip. Emitted:\n$folded"
        )
      case Left(err) =>
        fail(s"Failed to parse folded output: ${err.msg}\nEmitted:\n$folded")
    }
  }

  // -----------------------------------------------------------------------
  // Bug 7: Block scalars at top level emit at indent == 0.
  //
  // Per §8.1.1.1 and §6.1, block scalar content indentation level is
  // n + m, and the explicit indicator range is 1-9 (rule [163]).
  // At n=0, content must still be indented at least 1 space.
  // -----------------------------------------------------------------------

  test("bug7: literal block scalar at top level must indent content") {
    val input = "abc\ndef"
    val result = presentScalar(input, style = ScalarStyle.Literal)
    val parsed = result.as[String]
    parsed match {
      case Right(value) =>
        assertEquals(
          value,
          input,
          s"Literal scalar at top level did not round-trip. Emitted:\n$result"
        )
      case Left(err) =>
        fail(s"Failed to parse literal output at top level: ${err.msg}\nEmitted:\n$result")
    }
  }

  test("bug7: folded block scalar at top level must indent content") {
    val input = "abc\ndef\n"
    val result = presentScalar(input, style = ScalarStyle.Folded)
    val parsed = result.as[String]
    parsed match {
      case Right(value) =>
        assertEquals(
          value,
          input,
          s"Folded scalar at top level did not round-trip. Emitted:\n$result"
        )
      case Left(err) =>
        fail(s"Failed to parse folded output at top level: ${err.msg}\nEmitted:\n$result")
    }
  }

  // -----------------------------------------------------------------------
  // Bug 8: sys.error on non-scalar mapping keys is a regression.
  //
  // Per §3.2.1.1, "keys may be arbitrary nodes." Complex keys are legal
  // YAML and the `?` indicator exists for emitting them. Throwing a bare
  // RuntimeException is non-compliant.
  // -----------------------------------------------------------------------

  test("bug8: non-scalar mapping key should not throw") {
    val events = Seq(
      DocumentStart(),
      MappingStart(),
      MappingStart(),
      Scalar("nested", metadata = NodeEventMetadata(tag = Some(Tag.str))),
      Scalar("key", metadata = NodeEventMetadata(tag = Some(Tag.str))),
      MappingEnd,
      Scalar("value", metadata = NodeEventMetadata(tag = Some(Tag.str))),
      MappingEnd,
      DocumentEnd()
    )
    val result =
      try {
        val yaml = PresenterImpl.asString(events)
        Right(yaml)
      } catch {
        case e: RuntimeException => Left(e.getMessage)
      }
    assert(
      result.isRight,
      s"Non-scalar mapping key should not throw, but got: ${result.left.getOrElse("")}"
    )
  }

  // -----------------------------------------------------------------------
  // Roundtrip: per §3.2.1, §3.3.2, §10.3.2, a !!str node whose content
  // matches a non-string resolution pattern (null, true, 3.9, etc.) must
  // be quoted or explicitly tagged. Otherwise re-parsing resolves it to
  // the wrong type.
  // -----------------------------------------------------------------------

  test("roundtrip: string 'null' must survive encode-decode as string") {
    val yaml = "null".asYaml
    val parsed = yaml.as[String]
    parsed match {
      case Right(value) =>
        assertEquals(value, "null", s"String 'null' did not round-trip. YAML was: $yaml")
      case Left(err) =>
        fail(s"Failed to parse: ${err.msg}")
    }
  }
}
```

Please incorporate these tests and make them pass.

plokhotnyuk:
@lbialy Thanks for the review report!

I've withdrawn my attempt for folded and literal encodings and just focused to quoted and double-quoted ones.

Also, I've removed redundant whitespaces after ':' and '-' characters when no more values follow them.

plokhotnyuk:
@lbialy @tgodzik The PR is ready for review! Thank you!

lbialy:
native tests failed, I restarted them, I'll re-review the PR soon if it's green (should be I think)

lbialy:
@plokhotnyuk could you contribute your benchmarking setup for scala-yaml or at the very least leave a link to a github repo where it lives? I guess it would be good to have such a setup to track regressions before we release 1.0.0

btw.: I'm finishing review, brb in a moment

core/shared/src/main/scala/org/virtuslab/yaml/internal/dump/present/PresenterImpl.scala
```diff
@@ -1,102 +1,242 @@
...
+        sb.append(c)
+        i += 1
+      }
+      sb.append('\'')
+    }
+
+    def isCharNonPrintable(c: Char): Boolean =
+      c <= '\u001F' || // C0 control block
+        c == '\u007F' ||
+        c >= '\u0080' && c <= '\u009F' || // C1 control block (except for NEL \u0085)
```
lbialy: is this comment correct? the range includes `\u0085` and that's correct behavior, the comment seems wrong

core/shared/src/main/scala/org/virtuslab/yaml/YamlEncoder.scala
```diff
@@ -36,39 +36,37 @@ object YamlEncoder extends YamlEncoderCrossCompanionCompat {
...
 
-  implicit def forByte: YamlEncoder[Byte]       = v => Node.ScalarNode(v.toString)
-  implicit def forChar: YamlEncoder[Char]       = v => Node.ScalarNode(v.toString)
-  implicit def forShort: YamlEncoder[Short]     = v => Node.ScalarNode(v.toString)
-  implicit def forInt: YamlEncoder[Int]         = v => Node.ScalarNode(v.toString)
-  implicit def forLong: YamlEncoder[Long]       = v => Node.ScalarNode(v.toString)
-  implicit def forFloat: YamlEncoder[Float]     = v => Node.ScalarNode(v.toString)
-  implicit def forDouble: YamlEncoder[Double]   = v => Node.ScalarNode(v.toString)
-  implicit def forBoolean: YamlEncoder[Boolean] = v => Node.ScalarNode(v.toString)
-  implicit def forString: YamlEncoder[String]   = v => Node.ScalarNode(escapeSpecialCharacters(v))
```
lbialy, line 47: since this drops usage of `escapeSpecialCharacters` that method is dead code, it's not referenced anywhere else, remove it please

core/shared/src/test/scala/org/virtuslab/yaml/BaseYamlEncoderSuite.scala
```diff
@@ -0,0 +1,217 @@
...
+    val data = Seq("Mark McGwire", "Sammy Sosa", "Ken Griffey")
+    val expected =
+      s"""- Mark McGwire
+         |- Sammy Sosa
+         |- Ken Griffey
+         |""".stripMargin
+    assertEquals(data.asYaml, expected)
+  }
+
+  test("sequence of sequences") {
```
lbialy: this tests seems to be a duplicate of the same test above, was this meant to be a different test similar to the one above or is this just an error? if so, please delete

core/shared/src/main/scala/org/virtuslab/yaml/YamlEncoder.scala
```diff
@@ -36,39 +36,37 @@ object YamlEncoder extends YamlEncoderCrossCompanionCompat {
...
+  implicit def forByte: YamlEncoder[Byte]       = v => new Node.ScalarNode(v.toString, Tag.int)
+  implicit def forChar: YamlEncoder[Char]       = v => new Node.ScalarNode(v.toString, Tag.str)
```
lbialy, line 18: this is probably going to introduce breakage for users doing golden/frozen tests but it's a correct change, we should release this as 0.4.0 though, nothing to fix, just making notes :)

lbialy (changes requested):
I'm going to need to fix `SerializerImpl.convertScalarNode` when working on round trip fix and then fix your changes to work correctly in terms of the commented out tests but let's merge this as-is in regard to these Folded and Literal styles and these commented out tests

minor nitpicks left, also - we will need to release as 0.4.0 because we're breaking previous (incorrect) behavior

core/shared/src/main/scala/org/virtuslab/yaml/YamlEncoder.scala
plokhotnyuk: Yes, we should release 0.4.0 due to breaking changes and no backward binary compatibility due to removed public methods

core/shared/src/test/scala/org/virtuslab/yaml/BaseYamlEncoderSuite.scala
plokhotnyuk: Done

core/shared/src/main/scala/org/virtuslab/yaml/YamlEncoder.scala
plokhotnyuk: Done

core/shared/src/main/scala/org/virtuslab/yaml/internal/dump/present/PresenterImpl.scala
plokhotnyuk: Done

core/js/src/test/org/virtuslab/yaml/YamlEncoderSuite.scala
plokhotnyuk: Done

plokhotnyuk:
@lbialy Thanks for the review!

I will contribute benchmarks on the weekend

lbialy (approved):

